PR #39, opened 2021-02-16
Reformat pipeline.json to make it simpler in the UI

sxa:
Rework of https://github.com/AdoptOpenJDK/openjdk-build/pull/2394 which makes it easier to read the JSON in the jenkins UI when you want to add/remove variants - JDK11 only for now to verify if it works well, then we can do the same for other releases


gdams (approved):

M-Davies (approved):

sxa:
Merging ... Worst case we can revert quickly ;-)

PR #67, opened 2021-03-01
Define available jenkins configuration parameters

M-Davies:
* Centralised location for groovy config param descriptions and usage
* Also deletes the FAQ.md of this repo as it's not being removed by https://github.com/AdoptOpenJDK/openjdk-build/pull/2463

Part 1 of https://github.com/AdoptOpenJDK/openjdk-build/issues/2129
Signed-off-by: Morgan Davies <morgandavies2020@gmail.com>

### DIFF
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -1,0 +1,21 @@
+###########################
+###########################
+## Markdown Linter rules ##
+###########################
+###########################
+
+# Linter rules doc:
+# - https://github.com/DavidAnson/markdownlint
+#
+# Note:
+# To comment out a single error:
+#   <!-- markdownlint-disable -->
+#   any violations you want
+#   <!-- markdownlint-restore -->
+#
+
+###############
+# Rules by id #
+###############
+MD013: false # Line length is usually not important and the 80 char limit is way too small anyway
+MD033: false # Inline HTML is important for multilines and checkboxes within markdown tables
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,4 +1,4 @@
-# openjdk-build FAQ
+# ci-jenkins-pipelines FAQ
 
 This document covers cover how to perform various repeatable tasks in the
 repository that might not otherwise be obvious from just looking at the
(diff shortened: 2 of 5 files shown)


sxa (commented):
I'm not sure about a full removal of the FAQ.md. It is very difficult for people to get up to speed in stuff in our repositories and a basic document of "To do this, change this file" without all the background infromation is extremely valuable (I made a similar comment in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1974#pullrequestreview-598344190). If you really think we have no such content left then I mighe be ok with that, but it seems unlikely ...

M-Davies:
@sxa The FAQ is still going to available in https://github.com/AdoptOpenJDK/openjdk-build, even after #2463. I'm just removing it from here so we don't have two duplicate FAQ's floating around

sxa:
> available

There is stuff that would be specific to this repository - in fact a lot of the "end-user" stuff including running `make-adopt-build-farm.sh` vs `makejdk-any-platform.sh` could be in one here and expecting people to go to a repository about the pipelines for that and stuff like the "find your way around" chart and compiler prereqs seems a little odd so the information should perhaps be split between both repositories too.

M-Davies:
> > available
> 
> There is stuff that would be specific to this repository - in fact a lot of the "end-user" stuff including running `make-adopt-build-farm.sh` vs `makejdk-any-platform.sh` could be in one here and expecting people to go to a repository about the pipelines for that and stuff like the "find your way around" chart and compiler prereqs seems a little odd so the information should perhaps be split between both repositories too.

In that case, would you prefer two FAQ's? One here and one in openjdk-build, each answering questions about their respective repositories?

sxa:
> > > available
> > 
> > 
> > There is stuff that would be specific to this repository - in fact a lot of the "end-user" stuff including running `make-adopt-build-farm.sh` vs `makejdk-any-platform.sh` could be in one here and expecting people to go to a repository about the pipelines for that and stuff like the "find your way around" chart and compiler prereqs seems a little odd so the information should perhaps be split between both repositories too.
> 
> In that case, would you prefer two FAQ's? One here and one in openjdk-build, each answering questions about their respective repositories?

I think that's the only sensible choice unless the infromation specific to this repository is made available elsewhere in here.

karianna (approved):
LGTM

M-Davies:
Twin PR for updating the FAQ of openjdk-build -> https://github.com/AdoptOpenJDK/openjdk-build/pull/2506

sxa (approved):
LGTM as far as I can tell - thanks :-)

PR #68, opened 2021-03-01
add support for building Bisheng JDK11 on RISC-V

sxa:
As far as I can tell the script changes that to enable Bisheng for (https://github.com/AdoptOpenJDK/openjdk-build/issues/2420) work ok (I've been able to build natively in about three hours, which is still acceptable compared to the time [other JDK11 pipelines take](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/) including testing cycles, so subject to @karianna 's concerns in https://github.com/AdoptOpenJDK/openjdk-build/pull/2485 I want to try prototyping this on the machines we have. If we go ahead with this in the short term we can move to cross-compiling if appropriate later.

### DIFF
--- a/pipelines/jobs/configurations/jdk11u.groovy
+++ b/pipelines/jobs/configurations/jdk11u.groovy
@@ -22,7 +22,7 @@ weekly_release_scmReferences=[
         "openj9"         : "",
         "corretto"       : "",
         "dragonwell"     : "",
-        "bisheng"	 : ""
+        "bisheng"        : ""
 ]
 
 return this
--- a/pipelines/jobs/configurations/jdk11u.groovy
+++ b/pipelines/jobs/configurations/jdk11u.groovy
@@ -8,7 +8,7 @@ targetConfigurations = [
         "s390xLinux"    : [	"hotspot",	"openj9"					],
         "aarch64Linux"  : [	"hotspot",	"openj9",	"dragonwell"			],
         "arm32Linux"    : [	"hotspot"							],
-        "riscv64Linux"  : [			"openj9"			"bisheng"	]
+        "riscv64Linux"  : [			"openj9",			"bisheng"	]
 ]
 
 // 18:05 Tue, Thur
--- a/pipelines/jobs/configurations/jdk11u.groovy
+++ b/pipelines/jobs/configurations/jdk11u.groovy
@@ -1,13 +1,13 @@
 targetConfigurations = [
-        "x64Mac"        : [	"hotspot",	"openj9"					],
-        "x64Linux"      : [	"hotspot",	"openj9",	"dragonwell",	"corretto"	],
-        "x64Windows"    : [	"hotspot",	"openj9",	"dragonwell"			],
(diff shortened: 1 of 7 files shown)


sxa:
I'm fairly sure this is ok now ... Ideally needs https://github.com/AdoptOpenJDK/openjdk-build/pull/2509 to be merged first though

sxa:
run tests

sxa:
Notes for reviewers: We previously only had an OpenJ9 JDK available for use as both the bootstrap and the jenkins agent. That version currently does not have a JIT and has two issues. Firstly it takes about 7 hours to build, and secondly it takes several hours to archive the output and hits the 3 hour timeout from that phase. With the bisheng JVM used as bootstrap and jenkins agent VM, that drops to about 90 minutes, which is still quicker than the arm32 builds - at least until we get out bank of pis running. This means that in terms of adding it to the build pipelines without impact, we do not require this to be cross-compiled.

gdams (approved):

pipelines/jobs/configurations/jdk11u.groovy
```diff
@@ -21,7 +21,8 @@ weekly_release_scmReferences=[
         "hotspot"        : "",
         "openj9"         : "",
         "corretto"       : "",
-        "dragonwell"     : ""
+        "dragonwell"     : "",
+        "bisheng"	 : ""
```
M-Davies, line 25: Do we want this lined up?

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
```diff
@@ -172,12 +172,20 @@ class Config11 {
...
-        ],
+                crossCompile      : [
+                        "openj9"     : 'x64'
+                ],
+                buildArgs         : [
+                        "openj9"     : '--cross-compile'
+                ],
+                configureArgs     : [
+                        "openj9"     : '--disable-ddr --openjdk-target=riscv64-unknown-linux-gnu --with-sysroot=/opt/fedora28_riscv_root'
+                ]
```
M-Davies, line 187: And these?

M-Davies (approved):

pipelines/jobs/configurations/jdk11u.groovy
sxa: Yeah - the online merge conflict resolution wasn't doing a good job of showing me the spacing properly as I recall ...

sxa: Done - it was tab characters that messed with it - replaced with spaces now

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
sxa: In this case it was a bit of a cut & paste from elsewhere in the file where they weren't lined up bit I'll fix that too.

sxa:
If no-one spots any more glaring errors by the time the checks finish I'll merge this

PR #75, opened 2021-03-04
Auto-generate test jobs

### LINKED ISSUE #58: Auto-generate test jobs with a larger set of parameters
In [openjdk_build_pipeline.groovy/runTests](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L235), we [look for pre-created test job](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L269) names, run tests if we find them, and [print warning](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L285) if not.  This check could now be changed to check if job is not found, _generate_ it, then proceed to executing the test job.

Some options for approach are:

1. Grab the test job dsl file ([testJobTemplate](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate)) like how we [grab a single file from a separate repo in this getDependency](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/getDependency#L38) example and use it to generate jobs when they do not exist via the same approach as in Jenkinsfilebase (and how build jobs are generated)
1. Or run the [Test_Job_Auto_Gen](https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_Job_Auto_Gen/) job to generate jobs, similar to the [OpenJ9 approach](https://github.com/eclipse/openj9/blob/a685bffd936bd787b3a6ec6cf247cfad571c9a5c/buildenv/jenkins/common/pipeline-functions.groovy#L610) 

The expected pattern is that at the time that a new JDK version is beginning to be built, a bunch of test jobs get created on that first attempt to run testing.  If we take option 1, it does not serve to refresh the test jobs, we do not regenerate each time we run them (which is perfectly fine, as we can trigger Test_Job_Auto_Gen directly for that purpose).  

We had previously kept this auto-gen of test jobs a manual process until now partly to understand what new test jobs were being requested.  Hopefully, by adding this auto-gen of test jobs will not further faciliate the lack of communication and awareness between build/test components, but still, now is a good time to introduce it.

This issue is related to other enhancements:
- expanding the set of test job parameters that can be set (needed by SmokeTestsFoBuildAndPackaging below to pass VENDOR_REPO/BRANCH and requested to add LABEL by https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/issues/24)
- calling [SmokeTestsFoBuildAndPackaging](https://ci.adoptopenjdk.net/view/work-in-progress/job/SmokeTestsForBuildandPackaging/) from build pipelines (before any other tests are run, failure of this job should block all other testing) - see https://github.com/AdoptOpenJDK/openjdk-tests/issues/2024
- updating testJobTemplate to generate a full set of test job parameters divided into logical groups with parameter separator - see https://github.com/AdoptOpenJDK/openjdk-tests/issues/2238

sophia-guo:
Auto-generate test jobs with a larger set of parameters if test jobs
doesn't exist

Close #58 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -132,33 +132,39 @@ class Build {
     The test jobs all follow the same name naming pattern that is defined in the openjdk-tests repository.
     E.g. Test_openjdk11_hs_sanity.system_ppc64_aix
     */
-    def determineTestJobName(testType) {
-
-        def variant
+    def getTestJobParams(testType) {
+        def jobParams = [:]
+        def (level, group) = testType.tokenize('.')
+        jobParams.put('LEVELS', level)
+        jobParams.put('GROUPS', group)
         def number = getJavaVersionNumber()
-
+        jobParams.put('JDK_VERSIONS', number)
+        def variant
         switch (buildConfig.VARIANT) {
             case "openj9": variant = "j9"; break
             case "corretto": variant = "corretto"; break
             case "dragonwell": variant = "dragonwell"; break;
             default: variant = "hs"
         }
-
+        jobParams.put('JDK_IMPL', variant)
         def arch = buildConfig.ARCHITECTURE
         if (arch == "x64") {
             arch = "x86-64"
         }
-
-        def os = buildConfig.TARGET_OS
-
-        def jobName = "Test_openjdk${number}_${variant}_${testType}_${arch}_${os}"
-
+        def arch_os = "${arch}_${buildConfig.TARGET_OS}"    
+        def jobName = "Test_openjdk${number}_${variant}_${testType}_${arch_os}"
         if (buildConfig.ADDITIONAL_FILE_NAME_TAG) {
             switch (buildConfig.ADDITIONAL_FILE_NAME_TAG) {
-                case ~/.*XL.*/: jobName += "_xl"; break
+                case ~/.*XL.*/: 
+                    jobName += "_xl";
+                    arch_os += "_xl";
+                    break
             }
         }
-        return "${jobName}"
+        jobParams.put('TEST_JOB_NAME', jobName)
+        jobParams.put('ARCH_OS_LIST', arch_os)
+        jobParams.put('LIGHT_WEIGHT_CHECKOUT', true)
+        return jobParams
     }
 
     /*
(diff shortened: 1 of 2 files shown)


sophia-guo:
run tests

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -259,29 +265,30 @@ class Build {
...
-                                                context.string(name: 'JDK_BRANCH', value: jdkBranch),
-                                                context.string(name: 'OPENJ9_BRANCH', value: openj9Branch),
-                                                context.string(name: 'LABEL_ADDITION', value: additionalTestLabel),
-                                                context.string(name: 'KEEP_REPORTDIR', value: "${keep_test_reportdir}"),
-                                                context.string(name: 'ACTIVE_NODE_TIMEOUT', value: "${buildConfig.ACTIVE_NODE_TIMEOUT}")]
-                            }
-                        } else {
-                            context.println "[WARNING] Requested test job that does not exist or is disabled: ${jobName}"
+                        // Create test job if job doesn't exit or is not runnable
+                        if (!JobHelper.jobIsRunnable(jobName as String)) {
```
M-Davies, line 273: Can include a console out warning here that we are creating the jobs since they don't exist (easier for triagers to see that the job has just been created)

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -259,29 +265,30 @@ class Build {
...
+                            context.sh('curl -Os https://raw.githubusercontent.com/AdoptOpenJDK/openjdk-tests/master/buildenv/jenkins/testJobTemplate')
+							def templatePath = 'testJobTemplate'
+							jobDsl targets: templatePath, ignoreExisting: false, additionalParameters: jobParams
```
M-Davies, line 276: Indentation/whitespace isn't lined up

M-Davies (changes requested):
A couple of little points for quality of life 🙂 

Can we also update the [regeneration documentation](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/blob/master/pipelines/build/regeneration/README.md#downstream-test-jobs) to reflect these changes?

sophia-guo:
@M-Davies Thanks for reviewing. Updated.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -259,29 +265,31 @@ class Build {
...
-                                                context.string(name: 'JDK_REPO', value: jdkRepo),
-                                                context.string(name: 'JDK_BRANCH', value: jdkBranch),
-                                                context.string(name: 'OPENJ9_BRANCH', value: openj9Branch),
-                                                context.string(name: 'LABEL_ADDITION', value: additionalTestLabel),
-                                                context.string(name: 'KEEP_REPORTDIR', value: "${keep_test_reportdir}"),
-                                                context.string(name: 'ACTIVE_NODE_TIMEOUT', value: "${buildConfig.ACTIVE_NODE_TIMEOUT}")]
-                            }
-                        } else {
-                            context.println "[WARNING] Requested test job that does not exist or is disabled: ${jobName}"
+                        // Create test job if job doesn't exit or is not runnable
```
smlambert, line 272: minor typo: exit -> exist

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -132,33 +132,39 @@ class Build {
...
-        def os = buildConfig.TARGET_OS
-
-        def jobName = "Test_openjdk${number}_${variant}_${testType}_${arch}_${os}"
-
+        def arch_os = "${arch}_${buildConfig.TARGET_OS}"    
+        def jobName = "Test_openjdk${number}_${variant}_${testType}_${arch_os}"
         if (buildConfig.ADDITIONAL_FILE_NAME_TAG) {
             switch (buildConfig.ADDITIONAL_FILE_NAME_TAG) {
-                case ~/.*XL.*/: jobName += "_xl"; break
+                case ~/.*XL.*/:
```
smlambert, line 158: And eventually with the move to mixedref builds we may not need this case at all, but for now we will still keep it.  :)

smlambert (approved):
LGTM (minor typo in comment aside)

pipelines/build/common/openjdk_build_pipeline.groovy
sophia-guo: Updated. Thanks.

M-Davies (approved):
Thanks! LGTM now :)

smlambert:
hi @sophia-guo, I will merge this now.  I've just realized once its merged, then we'll want a follow-up PR... 

One PR in particular will be to correct the default list... right now `sanity.external` would try to run on windows/AIX and other platforms that do not have sw.tool.docker LABEL, and currently they don't actually try to run (and then would fail) because there is no job generated for them.  Once this is in, a job will get generated, it will try to run and hang looking for machines.

sophia-guo:
@smlambert I will do the follow up PR now if no body is working on it. Thanks Shelley!

PR #76, opened 2021-03-04
actions: add label bot

gdams:


M-Davies (changes requested):
I am very confused on what this is doing and I imagine anyone looking at the code without this commit reference title will be even worse off. Could we add in some documentation (and/or a link to it in the comments) explaining what is going on here?

gdams:
@M-Davies updated PTAL. I figured that inline docs are likely most appropriate here.

PR #78, opened 2021-03-04
Enable testing of OpenJDK 16/17 builds for Alpine

aahlenst:


gdams:
@aahlenst how do you plan to run the tests? We currently have no static alpine test agents connected to jenkins? The obvious approach here was to use the docker setup that we have for the build but I don't think it's been ported to the tests yet

smlambert:
I think we have 2 test agents now (via https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1929#issuecomment-783465921) and I have added test support via https://github.com/AdoptOpenJDK/openjdk-tests/pull/2304 (systemtest will fail to compile native tests at the moment, but all other default testing will run).

smlambert (approved):

smlambert:
Addresses part of https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/issues/70

gdams (approved):

PR #85, opened 2021-03-05
Disable external tests as weekly testlist for now

### LINKED ISSUE #70: Update nightly and weekly test lists
As we have enabled testing for jdk16 and jdk17 and adding testing for some new platforms (see https://github.com/AdoptOpenJDK/openjdk-tests/pull/2304), revisit and adjust nightly and weekly test lists to suit the number of test jobs running on the set of test machines available.

Fix the case that sanity.external is being launched against some platforms that it should not be on (remove it from [here](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/blob/master/pipelines/build/common/build_base_file.groovy#L87)), and instead of subtracting it from corretto (in this [file](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L244)), add it for hotspot/openj9.

sophia-guo:
Close #70 

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -83,8 +83,7 @@ class Builder implements Serializable {
     final List<String> weekly = [
         'extended.openjdk',
         'extended.perf',
-        'special.functional',
-        'sanity.external'
+        'special.functional'
     ]
 
     // Declare timeouts for each critical stage (unit is HOURS)
@@ -134,7 +133,7 @@ class Builder implements Serializable {
             buildArgs += " " + additionalBuildArgs
         }
 
-        def testList = getTestList(platformConfig, variant)
+        def testList = getTestList(platformConfig)
 
         def platformCleanWorkspaceAfterBuild = getCleanWorkspaceAfterBuild(platformConfig)
 
@@ -215,7 +214,7 @@ class Builder implements Serializable {
     Get the list of tests to run from the build configurations.
     We run different test categories depending on if this build is a release or nightly. This function parses and applies this to the individual build config.
     */
-    List<String> getTestList(Map<String, ?> configuration, variant) {
+    List<String> getTestList(Map<String, ?> configuration) {
         List<String> testList = []
         /*
         * No test key or key value is test: false  --- test disabled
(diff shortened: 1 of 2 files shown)


pipelines/build/common/build_base_file.groovy
```diff
@@ -247,8 +247,12 @@ class Builder implements Serializable {
 
             }
         }
-
+        
         testList.unique()
+        
+        if (testList.contains('sanity.external') && configuration.arch != 'ppc64le' && configuration.arch != 's390x' && !(configuration.os == 'linux' && configuration.arch == 'x64')) {
+            testList.remove(testList.indexOf('sanity.external'))
+        }
```
M-Davies, line 255: This is quite long conditional. Could we perhaps add a comment explaining why this is necessary with a link to the relevant issue?

M-Davies (changes requested):

M-Davies:
@sophia-guo You can tell if a build is running inside a docker container or not by checking if it has the `dockerImage` param set in the [build configs](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/blob/master/pipelines/jobs/configurations/jdk8u_pipeline_config.groovy#L26). See https://github.com/AdoptOpenJDK/ci-jenkins-pipelines#data-fields

smlambert (commented):
@sophia-guo - you are too fast !  

I have thought more about this, see my note in https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/issues/70#issuecomment-791545028, I think its fair for us to completely remove sanity.external from the list for now (we can keep running these tests like we do for extended.external separate from build pipelines on a schedule while we clean them up and land some enhancements like auto-update of tag-versions and possibly enabling running them outside of Docker / where they currently require test machines with sw.tool.docker (because external tests currently autogenerate a Dockerfile, build a Docker image and run the tests inside of it).

So, noting for @M-Davies benefit, external tests need to run on machines with the sw.tool.docker label, as in, Jenkins agents with Docker installed on them (which is different than Jenkins agents that are themselves docker agents).

sophia-guo:
@smlambert Thanks for explanation. Updated.

smlambert (approved):
thanks @sophia-guo !

M-Davies (approved):

PR #92, opened 2021-03-11
Inhibit nightly testing during JDK16 release cycle

sxa:
Title says it all - part of the standard release process.

### DIFF
--- a/pipelines/build/common/config_regeneration.groovy
+++ b/pipelines/build/common/config_regeneration.groovy
@@ -387,7 +387,7 @@ class Regeneration implements Serializable {
                 RELEASE: false,
                 PUBLISH_NAME: "",
                 ADOPT_BUILD_NUMBER: "",
-                ENABLE_TESTS: true,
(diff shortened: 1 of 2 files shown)


sxa:
@tsc Approval to merge during the lockdown cycle please

M-Davies (approved):
LGTM for actually disabling the tests (the affected jobs might require a script approval but hopefully not 🤞)

AdamBrousseau:
What's the best way for third parties to not pickup these types of temporary changes?
It looks like I could make copies and update my defaults.json. That runs the risk of my copies becoming stale.

smlambert (approved):

smlambert:
👍 for disabling during release prep

sxa:
> What's the best way for third parties to not pickup these types of temporary changes?

Suggest a separate issue covering that. I don't have an answer for it - it's been part of our release process for some time now (Although of others are using and shipping on the same cycles it may be worth considering doing the same).

PR #104, opened 2021-03-19
Always pull Docker image

AdamBrousseau:
Existing behaviour is to only pull DOCKER_IMAGE when
using the image as is. If the DOCKER_FILE path is
traversed, the base DOCKER_IMAGE is not refreshed
if it exists on the system. Fix is to move the pull
before the 'if' so that it is run for both cases.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1230,6 +1230,15 @@ class Build {
 
                             }
 
+                            // Pull the docker image from DockerHub
+                            try {
+                                context.timeout(time: buildTimeouts.DOCKER_PULL_TIMEOUT, unit: "HOURS") {
+                                    context.docker.image(buildConfig.DOCKER_IMAGE).pull()
+                                }
+                            } catch (FlowInterruptedException e) {
+                                throw new Exception("[ERROR] Master docker image pull timeout (${buildTimeouts.DOCKER_PULL_TIMEOUT} HOURS) has been reached. Exiting...")
+                            }
+
                             // Use our docker file if DOCKER_FILE is defined
                             if (buildConfig.DOCKER_FILE) {
                                 try {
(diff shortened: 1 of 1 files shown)


AdamBrousseau:
Tested both code paths on my farm, working as expected.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1257,13 +1265,6 @@ class Build {
 
                             // Otherwise, pull the docker image from DockerHub
```
M-Davies, line 1266: Now an irrelevant comment

M-Davies (changes requested):
With this change, does this invalidate the need for the `if` conditional determining if we should check for a `buildConfig.DOCKER_FILE`? If we are always going to pull, do we need to have the logic to build? I am unsure but do we have some images that we always need to build that are not stored on dockerhub? If so, this could be a breaking change for those builds

pipelines/build/common/openjdk_build_pipeline.groovy
AdamBrousseau: Moved & adjusted.

AdamBrousseau:
@M-Davies I can think of 3 use cases but only the first 2 seem to be supported currently
1. Docker image provided
2. Docker image & Docker file provided
3. Docker file provided
 
Based on L1199 the code is written that you must at least supply the image. The use case for a file is when you build an image during the build with a base image `DOCKER_IMAGE`. This is what the OpenJ9 builds do with the addition of cuda". You could build these and store them, eliminating the need to do it in the build. That's a separate topic. But I think the builds should also support the 3rd case of building given only a file which it doesn't appear they do at the moment. But that's also a separate discussion imo. I don't think this PR affects the use cases in any way.

M-Davies (approved):
Fair enough and [checking the docs, it doesn't look like anything needs to be updated over there](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines#data-fields). Let's give it a try

karianna (approved):
LGTM

PR #142, opened 2021-05-05
Run tests with PARALLEL=Dynamic

### LINKED ISSUE #109: Trigger AQA tests with PARALLEL=Dynamic
Given we have some large test targets that take a long time to complete (in particular extended.openjdk), AND we now have a good number of test machines for most platforms, it is time to shift to running tests with PARALLEL=Dynamic.

Switching to this approach will help break down the very large serially run test targets into smaller sets which will mean less test jobs that get aborted at the default timeout of 10hrs (which is happening frequently for extended.openjdk jobs on some platforms).

We already use this parallel mechanism at other Jenkins servers and for running in Azure Devops, so its well-exercised.  Caveats for use on our Jenkins server may be that we need to watch that the numbers of test artifacts being sent back to Jenkins master can be handled appropriately (though each artifact itself will be smaller).  We do not anticipate issues, but something to be aware of.

sophia-guo:
Close #109 


sophia-guo:
The initial change is to do some proof of concept. Not ready for review.

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -335,13 +335,18 @@ class Build {
                         }
 
                         def jobParams = getAQATestJobParams(testType)
+
+                        if ("${testType}" == "extended.openjdk") {
+                            jobParams.put("PARALLEL_DEFAULT", "Dynamic")
+                            jobParams.put("NUM_MACHINES", "2")
+                        }
```
AdamBrousseau, line 342: I know this is only a draft but just wanted to ensure that the final version of this is done via config file(s).
eg [defauls.json](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/defaults.json) and maybe also overwirtable on a per spec basis in [jdk11u_pipeline_config.groovy](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy)

sophia-guo: I believe there will be discussion about the levels of parallel for different test category. If it will eventually same for all different platform, variant, version, or it varies per platform, variant, version.  The final design will based on that this hack is just to test if there is anything unexpected for dynamic.

AdamBrousseau: Right, thanks, Sophia. As long as it is configurable so that different farms can set different values.

sophia-guo: Correct. Thanks for suggestion, Adam.

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

sophia-guo:
run tests

pipelines/build/common/build_base_file.groovy
```diff
@@ -256,7 +262,30 @@ class Builder implements Serializable {
 
         return testList
     }
-
+    /*
+    Get the list of tests to dynamically run  parallel builds from the build configurations.
+    This function parses and applies this to the individual build config.
+    */
+    Map<String, ?> getDynamicParas(Map<String, ?> configuration, String variant) {
```
smlambert, line 269: getDynamicParams would be easier to read than getDynamicParas (but I won't be a stickler on it)

pipelines/build/common/create_job_from_template.groovy
```diff
@@ -74,6 +74,8 @@ pipelineJob("$buildFolder/$JOB_NAME") {
                 <dt><strong>VARIANT</strong></dt><dd>hotspot, openj9...</dd>
                 <dt><strong>JAVA_TO_BUILD</strong></dt><dd>i.e jdk11u, jdk12u...</dd>
                 <dt><strong>TEST_LIST</strong></dt><dd>Comma seperated list of tests, i.e: sanity.openjdk,sanity.perf,sanity.system</dd>
+                <dt><strong>DYNAMIC_LIST</strong></dt><dd>Comma seperated list of tests, i.e: sanity.openjdk,sanity.perf,sanity.system</dd>
```
smlambert, line 77: minor spelling seperated should be spelled separated (line76 and 77)

smlambert (approved):
Only minor comments that I would not block on, thanks @sophia-guo !  (also, noting for other reviewers there is a quick overview of this work at the start of the AQAvit recording here: https://github.com/adoptium/aqa-tests/issues/2566#issuecomment-844570363).

andrew-m-leonard (approved):
lgtm

PR #152, opened 2021-05-26
Add dependency verification and lock files

johnoliver:


karianna (approved):
LGTM

AdamBrousseau (commented):
I noticed one the dependencies is jenkins-core:2.85. What are the implications of this? Does it mean that the build will not work on other version of Jenkins? Is there an issue with some background info regarding this change?

johnoliver:
@AdamBrousseau if jenkins make a breaking change then yes, but this PR is not setting this version, that value was set some time ago at: https://github.com/adoptium/ci-jenkins-pipelines/blob/750ea9bc326821b443645f6195d5aacc0a812f88/pipelines/build.gradle#L38

andrew-m-leonard:
this has been stale for over a year, converting to "draft", if no longer needed please close.

PR #23, opened 2021-02-10
Disable OpenJ9 XL Specs

AdamBrousseau:
Turn off large heap platforms in the ci builds.
More cleanup can be done related to xl packages
once the dust settles.

Depends #22

Signed-off-by: Adam Brousseau <adam.brousseau@ca.ibm.com>

PR #163, opened 2021-06-07
Openj9 XL spec cleanup

sophia-guo:
Related #23


sophia-guo:
@AdamBrousseau is it a good time for this clean up? Thanks.

AdamBrousseau (approved):
I think this is fine. I'm not sure what estimatedKey is used for to be honest and I've never liked how additionalFilename param was handled (hardcoded to xl). I don't believe the filename addition is used for anything else. I think we can leave the param but remove the xl code for it as you have done.

karianna (approved):
LGTM

gdams (approved):

PR #175, opened 2021-07-02
Enabling Solaris testing in JDK8 build pipeline

adamfarley:
We should be able to run these tests now. Adding them into the
automated pipeline.

Signed-off-by: Adam Farley <adfarley@redhat.com>

### DIFF
--- a/pipelines/jobs/configurations/jdk8u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk8u_pipeline_config.groovy
@@ -65,13 +65,13 @@ class Config8 {
         sparcv9Solaris: [
                 os  : 'solaris',
                 arch: 'sparcv9',
-                test: false
+                test: 'default'
         ],
 
         x64Solaris    : [
                 os                  : 'solaris',
                 arch                : 'x64',
-                test                : false
+                test                : 'default'
         ],
 
         ppc64leLinux  : [


adamfarley:
It seems like only these two single-line changes are required to enable testing for this, as the test jobs should be automatically generated at the end of the build job.

Second opinions welcome. Could I be missing something?

smlambert:
Looks correct.  To note that 'default' means that nightly default is run nightly and nightly + weekly is run on weekend & releases, as defined in the [defaults.json](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/defaults.json#L37-L49) file.  

So, the only other thing to watch out for would be do we have enough machine resource to cover that amount of testing.

adamfarley:
> 
> 
> Looks correct. To note that 'default' means that nightly default is run nightly and nightly + weekly is run on weekend & releases, as defined in the [defaults.json](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/defaults.json#L37-L49) file.
> 
> So, the only other thing to watch out for would be do we have enough machine resource to cover that amount of testing.

My theory was that, since we're only testing Hotspot JDK8 on these platforms, there's nothing else competing for the available machines (except for the JDK11 sparcv9Solaris builds, which don't run tests).

48 hours feels like more than enough time, given the run time of the default test sets on other platforms.

e.g. jdk8 on linux builds in 30 mins and takes 461 minutes to test (nightly). That's 7 hrs, 11 mins.

Solaris x64 builds in 20 mins, so assuming proportional run time for tests, just one machine should chew through all the tests in around 5 hrs.

Solaris sparcv9, however, takes 1 hr and 30 mins (ish) to build. Proportional testing times gives us an assumption of around 21 hrs for testing. Running once every 2 days still gives us 48hrs of runtime though, so we should be safe.

So we're *probably* fine, but I should still check on test runtime come Monday morning.

smlambert:
Based on a quick look at: https://ci.adoptopenjdk.net/label/sw.os.sunos/
there appears to be 1 sparcv9_solaris machine for test and 0 x86_64_solaris machines for test.  Will suggest holding off on merging this until we actually have machines available for testing, alternately, the PR can be adjusted to reducing the amount of testing for sparcv9_solaris and not enabling for x86-64_solaris until more resource is available.

smlambert:
Thanks @sxa for adding an x86_solaris machine to the pool, given that, we can try this out and see if amount of testing needs adjustment/reduction, or if given its only for jdk8 we get through the set before the next nightly is triggered.

smlambert (approved):

sxa (approved):
I'm going to approve and merge this on the basis that the JDK8u tests have gone through OK (11 and latest still running, but this only affects 8, and I would like to get it merged for the nightly pipelines due to kick off soon.

PR #181, opened 2021-08-17
Only run linter for pull requests targeting the master branch

keithc-ca:
It isn't helpful to re-run it in forks, particularly where it will only fail if the master branch does not exist.


karianna (approved):
LGTM

sxa (approved):
If I understand correctly this is to avoid running it when pushing to master in your own fork? If that's the case I approve

gdams (approved):

keithc-ca:
With this change you could still trigger the linter to run on your fork, but only via a pull request.

PR #191, opened 2021-09-15
Ensure that Windows builds get built and tested on a ZGC supported version of the O/S

karianna:


karianna:
TODO:  Actually this needs to be Windows 10 that this is tested on. So holding off as I'm not sure we can build and test on that OS just yet.

PR #194, opened 2021-09-27
Tag static-libs artefact as such in metadata

jerboaa:
See also: https://github.com/adoptium/temurin-build/issues/2699

Should get out of draft once/if https://github.com/adoptium/temurin-build/pull/2725 gets merged.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -953,6 +953,8 @@ class Build {
                 type = "testimage"
             } else if (file.contains("-debugimage")) {
                 type = "debugimage"
+            } else if (file.contains("-static-libs")) {
+                type = "staticlibs"
             }
 
             String hash = context.sh(script: """\


jerboaa:
> Seems OK, but needs a full pipeline run

@karianna Thanks. Can you explain what a "full pipeline run" means? Is that needed before or after this is integrated?

karianna:
> > Seems OK, but needs a full pipeline run
> 
> @karianna Thanks. Can you explain what a "full pipeline run" means? Is that needed before or after this is integrated?

I think it can go after - it's not harmful per se

tellison (approved):

sxa:
> Seems OK, but needs a full pipeline run

@jerboaa I've just kicked off https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk11-pipeline/1677 so we'll be able to see the output of the jobs run from that run. If you can't see that first link, the individual jobs in https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/ are the ones that have the artefacts accessible in them at the end.

jerboaa:
@sxa Thanks very much!

PR #214, opened 2021-11-18
Revert "Disable nightly testing for the October 2021 releases"

andrew-m-leonard:
Reverts adoptium/ci-jenkins-pipelines#209


adamfarley (approved):
LGTM

sxa (approved):
We'll need to add a comment about the pipeline regen requirement into the [RELEASING.md](https://github.com/adoptium/temurin-build/blob/master/RELEASING.md) too so that part doesn't get missed during the cycle. Hadn't realised it wouldn't have taken effect for so long :-)

PR #218, opened 2021-12-06
Ensure MacOS signing jenkins context does not overwrite existing BUILD_ARGS

andrew-m-leonard:
The MacOS signing jenkins context pipeline environment was overwriting the BUILD_ARGS when setting --make-exploded-image.
This PR appends --make-exploded-image to the existing variable.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1192,7 +1192,7 @@ class Build {
                                 context.println "[CHECKOUT] Checking out to adoptium/temurin-build..."
                                 repoHandler.checkoutAdoptBuild(context)
                                 if (buildConfig.TARGET_OS == "mac" && buildConfig.JAVA_TO_BUILD != "jdk8u") {
-                                    context.withEnv(["BUILD_ARGS=--make-exploded-image"]) {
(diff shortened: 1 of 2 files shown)


andrew-m-leonard:
Test builds: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-test-mac/

andrew-m-leonard:
run tests

andrew-m-leonard:
Looks like PR tester is broken by enableTestDynamicParallel

smlambert (approved):
thanks @andrew-m-leonard !

andrew-m-leonard:
run tests

karianna:
Have we got this the wrong way around?  Should we _only_ allow the setting of BUILD_ARGS from external scripts as config?  This seems to be the only place we do this?

andrew-m-leonard:
> Have we got this the wrong way around? Should we _only_ allow the setting of BUILD_ARGS from external scripts as config? This seems to be the only place we do this?

@karianna i'd have to defer to @gdams, the mac assemble/exploded builds use both --make-exploded-image and --assemble-exploded-image separately, seems a MacOS special case, you wouldn't be able to specify them both in the BUILD_ARGS as config

gdams (approved):

PR #226, opened 2022-01-06
Create "temurin" variant distinct from "hotspot"

sxa:
The other half of https://github.com/adoptium/temurin-build/pull/2818

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ x64Mac        : [
     os                   : 'mac',
     arch                 : 'x64',
     additionalNodeLabels : [
-            hotspot  : 'macos10.14',
+            temurin  : 'macos10.14',
             corretto : 'build-macstadium-macos1010-1',
             openj9   : 'macos10.14'
     ],
@@ -46,7 +46,7 @@ NOTE: When the `type` field implies a map, the `String` key of the inner map is 
 
 ```groovy
                 additionalNodeLabels : [
-                        hotspot : 'xlc13&&aix710',
+                        temurin : 'xlc13&&aix710',
                         openj9 :  'xlc13&&aix715'
                 ],
 ```
@@ -86,46 +86,46 @@ A single `Map<String, Map<String, String>>` variable containing what platforms a
 ```groovy
 targetConfigurations = [
         "x64Mac"        : [
-                "hotspot",
+                "temurin",
                 "openj9"
         ],
         "x64Linux"      : [
-                "hotspot",
+                "temurin",
                 "openj9",
                 "corretto",
                 "dragonwell"
         ],
         "x32Windows"    : [
-                "hotspot",
+                "temurin",
                 "openj9"
         ],
         "x64Windows"    : [
-                "hotspot",
+                "temurin",
                 "openj9",
                 "dragonwell"
         ],
         "ppc64Aix"      : [
-                "hotspot",
+                "temurin",
                 "openj9"
         ],
         "ppc64leLinux"  : [
-                "hotspot",
+                "temurin",
                 "openj9"
         ],
         "s390xLinux"    : [
-                "hotspot",
+                "temurin",
                 "openj9"
         ],
         "aarch64Linux"  : [
-                "hotspot",
+                "temurin",
                 "openj9",
                 "dragonwell"
         ],
         "arm32Linux"  : [
-                "hotspot"
+                "temurin"
         ],
         "sparcv9Solaris": [
-                "hotspot"
+                "temurin"
         ]
 ]
 ```
@@ -153,7 +153,7 @@ Source control references (e.g. tags) to use in the weekly release builds
 
 ```groovy
 weekly_release_scmReferences = [
-        "hotspot"        : "jdk8u282-b07",
+        "temurin"        : "jdk8u282-b07",
         "openj9"         : "v0.24.0-release",
         "corretto"       : "",
         "dragonwell"     : ""
@@ -227,7 +227,7 @@ This tag identifies the architecture the JDK has been built on and it intended t
 ---
 
 - `variant:`
-Example values: [`hotspot`, `openj9`, `corretto`, `dragonwell`]
+Example values: ['hotpost, `temurin`, `openj9`, `corretto`, `dragonwell`]
 
 This tag identifies the JVM being used by the JDK, "dragonwell" itself is not a JVM but is currently considered a variant in its own right.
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -227,7 +227,7 @@ This tag identifies the architecture the JDK has been built on and it intended t
 ---
 
 - `variant:`
-Example values: [`hotspot`, `openj9`, `corretto`, `dragonwell`]
+Example values: ['hotpost, `temurin`, `openj9`, `corretto`, `dragonwell`]
```
sophia-guo, line 230: typo `hotspot`?

Question: hotspot variant is still available and what does it mean (openjdk)?

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -212,6 +212,8 @@ class Build {
             } else if (buildConfig.VARIANT == "openj9") {
                 jdkBranch = 'openj9'
             } else if (buildConfig.VARIANT == "hotspot"){
+                jdkBranch = 'master'
```
sophia-guo, line 215: Does this mean hotspot variant is 'openjdk'?

sophia-guo:
Will this change the value of System.getProperty('java.vm.name')?[ TKG is using the value to set impl ](https://github.com/adoptium/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L114-L128)

sophia-guo:
Test jobs are using buildConfig.VARIANT directly to set impl parameter. [jobParams.put('JDK_IMPL', buildConfig.VARIANT)](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L184)

README.md
sxa: It means plain openjdk without any "Temurin" "Eclipse" or "Adoptium" markings on it, or require the `_adopt` git tags when building a tagged version.

This will mean that people can use our scripts in a default way to build their own JDK more easily from a branch of openjdk. The current process trips people up when they're trying to build GA versions particularly becuse of the missing `_adopt` tags and also we don't really want people building their own versions that call themselves "Temurin" by default. Hope that clarifies things.

pipelines/build/common/openjdk_build_pipeline.groovy
sxa: I think this is ok as both temurin and hotspot would go into the `variant = "hs"` section which is correct. I do NOT plan to build `VARIANT=hotspot` in jenkins, so at the moment the `hs` test jobs would be exclusively for the `temurin` variant and when I was testing last week with this change the `hs` jobs got invoked correctly. We may wish to change that later of course but there should be no reason to change from `hs` just now :-)

sxa:
> Will this change the value of System.getProperty('java.vm.name')?[ TKG is using the value to set impl ](https://github.com/adoptium/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L114-L128)

If it does it's a bug :-) I've only been testing the build process for now and making sure the correct vendor fields get passed in. `vm.name` should definitely be `hotspot` and I'll make sure that's still the case before this goes in.

sxa:
> Will this change the value of System.getProperty('java.vm.name')?[ TKG is using the value to set impl ](https://github.com/adoptium/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L114-L128)

It's coming out as `OpenJDK 64-Bit Server VM` in Both cases. The differences in properties between `hotspot` and `temurin` with this change are as follows:
## JDK17
```
21,23c21,24
<     java.vendor = Undefined Vendor
<     java.vendor.url = https://openjdk.java.net/
<     java.vendor.url.bug = https://bugreport.java.com/bugreport/
---
>     java.vendor = Eclipse Adoptium
>     java.vendor.url = https://adoptium.net/
>     java.vendor.url.bug = https://github.com/adoptium/adoptium-support/issues
>     java.vendor.version = Temurin-17.0.1+12-202201101740
32c33
<     java.vm.vendor = Undefined Vendor
---
>     java.vm.vendor = Eclipse Adoptium
55,56c56,57
```
## JDK8
```
26,28c26,28
<     java.vendor = Oracle Corporation
<     java.vendor.url = http://java.oracle.com/
<     java.vendor.url.bug = http://bugreport.sun.com/bugreport/
---
>     java.vendor = Temurin
>     java.vendor.url = https://adoptium.net/
>     java.vendor.url.bug = https://github.com/adoptium/adoptium-support/issues
35c35
<     java.vm.vendor = Oracle Corporation
---
>     java.vm.vendor = Temurin
67,68c67,68
```

sxa:
> > Test jobs are using buildConfig.VARIANT directly to set impl parameter. [jobParams.put('JDK_IMPL', buildConfig.VARIANT)](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L184)
> 
> The jobs seemed to run through ok - possibly just by falling into "default" options. A quick solution to this would be to just override the value if it's temurin and use `hotspot` instead. How does that sound? We can look at doing a better split later (`VARIANT` probably shouldn't be used for this as has been discussed elsewhere but this is a tactical fix to try and avoid people from building their own thing called `Temurin` using our scripts :-) )

ping @sophia-guo - does this sound reasonable, or do you know if it would fall into a default option that does the right thing regardless?

sxa:
Linkter failures are coverd under https://github.com/adoptium/temurin-build/issues/2831

sophia-guo:
> override the value if it's temurin and use hotspot instead

@sxa that should be good.

karianna (approved):
I'm largely good with this as it's a good stepping stone to the variant vs vendor split.

sxa:
> I'm largely good with this as it's a good stepping stone to the variant vs vendor split.

Agreed. My feeling is that if we can at least prove we can make the split (Which we'll hopefully find out in the next couple of days when these go in!) we can have confidence that such a subsequent change will be relatively easy to do. I just didn't like the prospect of making too many changes all at once in here, particuarly when it's risky and may need to be backed out if something very obvious doesn't work.

sxa:
> > override the value if it's temurin and use hotspot instead
> @sxa that should be good.

Done in latest commit in this PR: https://github.com/adoptium/ci-jenkins-pipelines/pull/226/commits/a69df7a7e4eb99b46e36662bc66fc62495aa403c

andrew-m-leonard (approved):

PR #228, opened 2022-01-10
Remove RELEASE_TAG, which has been removed from downstream jobs

sophia-guo:
Related with https://github.com/adoptium/temurin-build/pull/2166/ 
https://github.com/adoptium/aqa-tests/pull/2035


karianna:
Hmm, is this not used by OpenJ9 / Semeru builds or even temurin builds with the _adopt suffix tag?

sophia-guo:
I believe it's not used by Openj9 / Semeru builds @llxia could you confirm that?

What is the builds with _adopt suffix tag? Are we still running them?

karianna:
> I believe it's not used by Openj9 / Semeru builds @llxia could you confirm that?
> 
> What is the builds with _adopt suffix tag? Are we still running them?

I think we can ignore the _adopt suffix tags as that's silently handled now CC @gdams

llxia:
> I believe it's not used by Openj9 / Semeru builds

RELEASE_TAG is removed in aqa-tests repo. I do not think it is used by Openj9 / Semeru builds. FYI @AdamBrousseau

sophia-guo:
@smlambert could you please take a review? Thanks

smlambert (approved):
LGTM, thanks @sophia-guo !

PR #241, opened 2022-02-15
Adjust output filename for temurin to still be called hotspot

sxa:
I hadn't compensated for the output filename in https://github.com/adoptium/ci-jenkins-pipelines/pull/226 - it would be undesirable to change it to `temurin` at the moment since it would be an externally visible change. This should switch it back to having `hotspot` in the filename

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1024,7 +1024,12 @@ class Build {
 
         javaToBuild = javaToBuild.toUpperCase()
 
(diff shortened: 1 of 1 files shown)


sxa:
NOTE: Fell foul of the [macos installer code](https://github.com/adoptium/installer/blob/db2a15336bea5f486771fbd4f26bf133eb80b0f0/pkgbuild/create-installer-mac.sh#L48) as it failed to have a suitable definition if neither `hotspot` or `openj9` was found in the tarball filename

andrew-m-leonard (approved):

smlambert (approved):
👍

PR #247, opened 2022-02-17
Disable compilation of kae feature on bishengjdk 8

eapenzhan:
Since continuous failure of compilation of kae feature, which relies on openssl 1.1.1.f, we decide to stop the compilation.
It could be opened again when openssl 1.1.1 is popular enough. Whatever, just stop it right now. Thanks.


karianna (approved):
lgtm

sxa (approved):
Happy to approve this, although the number of systems that don't have 1.1.1 will be getting quite low know. I guess anything based on CentOS7/RHEL7 will be those causing problems as they only have 1.0.2

PR #249, opened 2022-02-17
Remove scmReference for unused hotspot variant from weekly config

sxa:
Removed in our pipelines in preference for `temurin` so these options are no longer required as @sophia-guo pointed out in https://github.com/adoptium/temurin-build/issues/2671#issuecomment-1043157258


karianna (approved):
LGTM

PR #253, opened 2022-02-25
Update Windows and VS to levels that support ZGC

karianna:


gdams:
run tests

sxa (commented):
Bear in mind that we currently only have [one win2019 machine](https://ci.adoptopenjdk.net/computer/test-azure-win2019-x64-1/) and it's marked as a `test-` system, so we'd need to ensure we have a machine available for this before proceeding. The existing machine is also currently offline dur to low disk space.

karianna:
Setting this to draft as infra work will be required first.

PR #255, opened 2022-02-28
Trigger Test_Job_Auto_Gen to generate AQA test jobs

llxia:
In the case of AQA_AUTO_GEN is set to true, if Test_Job_Auto_Gen is runnable, trigger Test_Job_Auto_Gen to generate AQA test jobs. Otherwise, use testJobTemplate.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -193,7 +193,7 @@ class Build {
         }
         def arch_os = "${arch}_${buildConfig.TARGET_OS}"
         jobParams.put('ARCH_OS_LIST', arch_os)
-        jobParams.put('LIGHT_WEIGHT_CHECKOUT', true)
+        jobParams.put('LIGHT_WEIGHT_CHECKOUT', false)
         return jobParams
     }
     /*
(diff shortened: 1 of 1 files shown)


llxia:
FYI @smlambert @AdamBrousseau

smlambert (approved):
Thanks @llxia, LGTM (and I have verified that we have some values that will not be passed in via the top-level build job set as defaults in Test_Job_Auto_Gen).

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -193,7 +193,7 @@ class Build {
         }
         def arch_os = "${arch}_${buildConfig.TARGET_OS}"
         jobParams.put('ARCH_OS_LIST', arch_os)
-        jobParams.put('LIGHT_WEIGHT_CHECKOUT', true)
+        jobParams.put('LIGHT_WEIGHT_CHECKOUT', false)
```
sophia-guo, line 196: What's the reason to update to 'false'?

sophia-guo:
The benefit of using Test_Job_Auto_Gen job is to easily pick up the default values set in the Test_Job_Auto_Gen. However the job configuration of Test_Job_Auto_Gen is not source controlled. That probably means, first, different jenkins servers will need to setup the job with exact job name (hard coded), second, Test_Job_Auto_Gen jobs might need to have same configuration and third any job configuration changes might need to be monitored to update accordingly in this jenkins files.

pipelines/build/common/openjdk_build_pipeline.groovy
llxia: If we set `LIGHT_WEIGHT_CHECKOUT ` true, it means AQA repo and branch are hardcoded. It will take the variable that we set in the job parameters.

llxia:
>The benefit of using Test_Job_Auto_Gen job is to easily pick up the default values set in the Test_Job_Auto_Gen.

Just to be clear, the benefit of using Test_Job_Auto_Gen job is to store parameters that are specific to each vendor (i.e., different credentials, private repos,  artifactory server/dir, etc)

>  That probably means, first, different jenkins servers will need to setup the job with exact job name (hard coded),

No, Test_Job_Auto_Gen job is not required. If Test_Job_Auto_Gen does not exist, it will run testJobTemplate.

> second, Test_Job_Auto_Gen jobs might need to have same configuration

Can be more specific about the same configuration? Test_Job_Auto_Gen consumes parameters that are passed in from this script. Test_Job_Auto_Gen should only store some specific info.

> third any job configuration changes might need to be monitored to update accordingly in this jenkins files

I am not sure I understand this point. The original logic runs testJobTemplate file. And Test_Job_Auto_Gen runs testJobTemplate underneath. There is no different running Test_Job_Auto_Gen and testJobTemplate, except Test_Job_Auto_Gen is a real jenkins job that contains some vendor special parameters.

sophia-guo:
won't block this PR just put the notes which probably need to re think later.

sophia-guo (approved):

PR #259, opened 2022-03-02
Enable temurin variant for now

### LINKED ISSUE #258: Nightly publish job with slack notifications no longer runs
We used to have a job which posted slack messages to the build channel, detailing whether or not latest nightly builds got published to github. This no longer works. I'm guessing due to https://github.com/adoptium/temurin-build/issues/2671

**Expected behaviour:**
Publish stats posted to build slack channel.

**Observed behaviour:**
No publish stats.

**Any other comments:**
https://github.com/adoptium/ci-jenkins-pipelines/blob/master/tools/nightly_build_and_test_stats.groovy#L35 and the like would probably need to get updated. It used to have `VARIANT=temurin` now it's `VARIANT=hotspot`

sophia-guo:
Fix #258 

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -32,7 +32,10 @@ node ("master") {
   def testStats = []
 
   stage("getPipelineStatus") {
-    if (variant == "temurin") { // temurin only for now
+    if (variant == "temurin") {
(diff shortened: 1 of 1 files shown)


sxa (approved):
Switching the the variant in the call to the API makes sense.

smlambert (approved):

PR #263, opened 2022-03-09
Enable sanity.external on JDK11 xlinux

llxia:
- Enable sanity.external on JDK11 xlinux (weekly)
- Set DYNAMIC_COMPILE to true for functional and external testing

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -347,6 +347,11 @@ class Build {
                             keep_test_reportdir = "true"
                         }
 
+                        def DYNAMIC_COMPILE = false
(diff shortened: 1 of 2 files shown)


llxia:
@smlambert @sophia-guo please review. Thanks

sophia-guo:
This should be fine for sanity.external on JDK11 xlinux (weekly). Going forward we may enable sanity.external as nightly and extended.external as weekly for all linux and all version as default. In that case probably updating [defaults.json](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/defaults.json#L35-L49) is preferred.

sophia-guo (approved):

llxia:
Not sure if we can update defaults.json as not all platforms have docker support. We will need to evaluate per platform.

smlambert (approved):
thanks @llxia !

PR #264, opened 2022-03-09
Ensure correct variant used for querying Temurin build jobs

andrew-m-leonard:
getStatus method was changing variant from temurin to hotspot, thus the later build job query which searches based on "temurin" did not find any build jobs.

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -32,17 +32,18 @@ node ("master") {
   def testStats = []
 
   stage("getPipelineStatus") {
-    if (variant == "temurin") {
-        variant = "hotspot"
+    def apiVariant = variant
+    if (apiVariant == "temurin") {
+        apiVariant = "hotspot"
     }
-    if (variant == "hotspot") { // hotspot only for now
+    if (apiVariant == "hotspot") { // hotspot only for now
       // Determine nightly pipeline health by looking at published assets.
       // In particular, look at one data set for published binaries (Linux x64).
       // If no published assets happened the last 4 days, the nightly pipeline
       // is considered unhealthy.
       // TODO: account for disabled nightly pipelines
       featureReleases.each { featureRelease ->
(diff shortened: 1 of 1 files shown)


tools/nightly_build_and_test_stats.groovy
```diff
@@ -32,17 +32,18 @@ node ("master") {
   def testStats = []
 
   stage("getPipelineStatus") {
-    if (variant == "temurin") {
-        variant = "hotspot"
+    def apiVariant = variant
```
sxa, line 35: Can we make this `trssVariant` or `trssApiVariant` instead of `apiVariant` to disambiguiate it from the main adoptium API?

sxa (commented):
Comment on the variable name, since I had to think about it for a second before I realised it was the TRSS API :-)

tools/nightly_build_and_test_stats.groovy
andrew-m-leonard: it is the adoptium API..
```${apiUrl}/v3/assets```

andrew-m-leonard: the problem was this code (for the publish health status) was overriding the variant variable to hotspot...

andrew-m-leonard: @sxa

smlambert:
Note (in case it affects how you have to handle updating this script) that we chose to leave the test names unchanged with this hotspot -> temurin rename in the build pipelines, to avoid downstream disruption (which is why this script is still able to retrieve test job info).  If you are changing this script, there are temurin build names and temurin smoke test job names, but the main test pipelines continue to use the old naming schema.

andrew-m-leonard:
> Note (in case it affects how you have to handle updating this script) that we chose to leave the test names unchanged with this hotspot -> temurin rename in the build pipelines, to avoid downstream disruption (which is why this script is still able to retrieve test job info). If you are changing this script, there are temurin build names and temurin smoke test job names, but the main test pipelines continue to use the old naming schema.

Yep, I noticed that. This just fixes the "build job" query.

sxa (approved):

tools/nightly_build_and_test_stats.groovy
sxa: Thanks for the clarification. LGTM

karianna (approved):
LGTM

PR #267, opened 2022-03-09
Set DYNAMIC_COMPILE=true for smoke tests

llxia:

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -292,7 +292,7 @@ class Build {
                         context.jobDsl targets: templatePath, ignoreExisting: false, additionalParameters: jobParams
                     }
                 }
-               context.catchError {
+                context.catchError {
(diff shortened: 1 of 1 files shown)


llxia:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3823/

smlambert (approved):
We restrict the BUILD_LIST to functional/buildAndPackage folder already I think, but this is a good plan to be certain we restrict compilation.

sophia-guo:
Yes, the BUILD_LIST is set to functional/buildAndPackage as default when jobs are generated. 

As DYNAMIC_COMPILE is set as constant there is no need to pass it every time.  I would suggest to set it in the getSmokeTestJobParams() as the default value if we'd like to be certain for restrict compilation. Though the setting won't immediately effect until jobs are regenerated( we don't have force regeneration choice for now) 

Passing in as parameter will effect right away, but  looks like it won't help smoke tests for now.

gdams (approved):

PR #269, opened 2022-03-16
Disable nightly sanity testing on RISC-V for JDK11 as it slows pipeline completion

sxa:
Caused JDK11u pipeline to still be running this morning. Until we are able to automate dynamic parallel testing on sanity, this should be removed.

### DIFF
--- a/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
@@ -158,7 +158,7 @@ class Config11 {
                         "bisheng"    : '--openjdk-target=riscv64-unknown-linux-gnu --with-sysroot=/opt/fedora28_riscv_root --with-jvm-features=shenandoahgc'
                 ],
                 test                : [
-                        nightly: ['sanity.openjdk'],
+                        nightly: false,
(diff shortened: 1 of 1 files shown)


jerboaa (approved):

karianna:
Do we have an issue to fix/re-enable somewhere?

sxa:
> Do we have an issue to fix/re-enable somewhere?

I don't think so, but it can still be run manually as desired/required as we move this forward. I've had initial discussions with @smlambert regarding enable parallel builds on things other than `extended.openjdk` but given the priority of this platform it's very much on a "best-effort" basis.

smlambert:
I also think we should separate Temurin & the 'supported platforms' we release on into its own pipeline (that has priority), and all other platforms/impls separated out to a secondary pipeline with lower priority.  Secondary pipelines should not delay or disrupt our critical path activities.

sxa:
> Do we have an issue to fix/re-enable somewhere?

Covered in https://github.com/adoptium/infrastructure/issues/2639 which is slowly moving forward as a background task.

sxa:
This has been moved to a separate pipeline, therefore closing this as it's not interfering with anything important now.

PR #283, opened 2022-04-13
Force make to use SHELL=/bin/bash on Solaris

sxa:
Fixes https://github.com/adoptium/temurin-build/issues/2894

I believe this will set `BUILD_ARGS` to the right thing in the pipelines - https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-x64-hotspot/795/ seems to be running happily with this in place:

```
"BUILD_ARGS": "--make-args SHELL=/bin/bash",
```

### DIFF
--- a/pipelines/jobs/configurations/jdk8u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk8u_pipeline_config.groovy
@@ -77,13 +77,15 @@ class Config8 {
         sparcv9Solaris: [
                 os  : 'solaris',
                 arch: 'sparcv9',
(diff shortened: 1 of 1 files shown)


gdams:
Are we sure this is the most appropriate place to make this change? To me this seems like something that should likely be set as part of https://github.com/adoptium/temurin-build/blob/master/build-farm/platform-specific-configurations/solaris.sh

sxa:
Agreed - replaced with https://github.com/adoptium/temurin-build/pull/2897

PR #287, opened 2022-04-28
Don't add whitespace if no existing configure args

AdamBrousseau:


AdamBrousseau:
We had an odd bug on our builds where a single configure arg was prefixed with a whitespace. This caused some parsing of that arg to fail. This fix will only add the whitespace if args already exist.

zdtsw (commented):
LGTM

karianna (changes requested):
LGTM

karianna (approved):

sophia-guo (approved):

zdtsw (approved):

PR #288, opened 2022-05-05
Omit MacOS v10.11 from testing in JDK17+ builds

jdekonin:
Signed-off-by: Joe deKoning <joe_dekoning@ca.ibm.com>


AdamBrousseau (commented):
This looks fine to me. I'm not sure if Adopt has any 10.11 machines or labels their test machines to this extent. If they are not interested in this change we can take it directly to our fork. We will also need a change there anyways for the Certified Edition specs which aren't here.

karianna (approved):
LGTM

sxa (approved):
I'm ok with this, although we don't use that tag in the adopt CI so this will only be relevant for the OpenJ9 CI runs. I decomissioned our older macos machines a few weeks ago - we no longer have anything earlier than 10.14.

PR #292, opened 2022-05-10
feat: add support to use aarch64 Alpine linux docker image by Jenkins agents

zdtsw:
- on 11, 16, 17,18, 19
Fixes: https://github.com/adoptium/infrastructure/issues/2497

### DIFF
--- a/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
@@ -43,6 +43,14 @@ class Config11 {
                 test                : 'default'
         ],
 
+        aarch64AlpineLinux  : [
+                os                  : 'alpine-linux',
+                arch                : 'aarch64',
+                dockerImage         : 'adoptopenjdk/alpine3_build_image',
+                test                : 'default',
+                configureArgs       : '--disable-headful'
+        ],
+
         x64Windows: [
                 os                  : 'windows',
                 arch                : 'x64',
--- a/pipelines/jobs/configurations/jdk16u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk16u_pipeline_config.groovy
@@ -36,6 +36,14 @@ class Config16 {
                 configureArgs       : '--enable-headless-only=yes'
         ],
 
+        aarch64AlpineLinux  : [
+                os                  : 'alpine-linux',
+                arch                : 'aarch64',
+                dockerImage         : 'adoptopenjdk/alpine3_build_image',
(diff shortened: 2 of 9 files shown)


karianna (approved):
It looks correct, anyway to manually test before merging?

zdtsw:
> It looks correct, anyway to manually test before merging?

I assume we do not have a "staging" Jenkins controller can be used for offline test.  
In that case, to have a e2e test, need to someone to create a dummy Jenkins job similar to https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-alpine-linux-x64-temurin but for aarch64, but config Parameters to use [aarch64](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-alpine-linux-x64-temurin/37/parameters/) once the docker image is ready in dockerhub then run the job manually to see how it goes. 
If purely to test the docker image itself, (skip how jenkins etc) that has been [done](https://github.com/adoptium/infrastructure/issues/2497#issuecomment-1103482790)

zdtsw:
> LGTM although we only build 8, 11, 17, 18, 19 so the 16 change isn't needed (but it doesn't matter too much)

Yeah, I added for 16 because x64 has alpine docker configed there. 
maybe we can have another issue to cleanup some old pipeline config files if the versions aint supported any longer.

sxa (changes requested):
This is going to need boot JDKs for JDK8 and 11 to be in place - I have added this under https://github.com/adoptium/infrastructure/pull/2559 which is not merged yet.

I've tried building JDK8 and it doesn't seem to work so I suggest removing the changes to [jdk8u_pipeline_config.groovy](https://github.com/adoptium/ci-jenkins-pipelines/pull/292/files#diff-87a99e1c88644292b09604fc649491eeecb9d29140458b9742491b67fb669351) for now

```
/temurin-build/build-farm/workspace/build/src/hotspot/src/os_cpu/linux_aarch64/vm/os_linux_aarch64.cpp:77:10: fatal error: fpu_control.h: No such file or directory
   77 | #include <fpu_control.h>
      |          ^~~~~~~~~~~~~~~
Compiling /temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/gc_implementation/parallelScavenge/parMarkBitMap.cpp
compilation terminated.
```

sxa (approved):
I'm working on getting a suitable JDK19 boot JDK in place for this for the first cycle but the others should get most of the way through the process, subject to the potential hang on JDK11 so I'm going to merge and see how it goes ...

PR #293, opened 2022-05-10
AIX: Use xlc16 compiler for JDK11

sxa:
To sync with https://github.com/adoptium/temurin-build/pull/2886

@AdamBrousseau I presume you're OK with using XLC16 for OpenJ9 JDK11? Would be good to get the OpenJ9 team's OK on this since the PR changes hotspot and openj9


sxa:
This is going to need FixPack11 to be installed on the XLC16 compiler to work - I'll trial this on one of the build machines...

andrew-m-leonard (approved):

PR #296, opened 2022-05-12
feat: enable --create-sbom (in temurin-build) by passing new BUILD_ARGS

zdtsw:
- by default it is set to "true"
	- should be able to turn it off per each groovy config file when generate pipeline

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -172,6 +172,7 @@ class Builder implements Serializable {
             ENABLE_TESTDYNAMICPARALLEL: enableTestDynamicParallel,
             ENABLE_INSTALLERS: enableInstallers,
             ENABLE_SIGNER: enableSigner,
+            ENABLE_SBOM: enableSBOM,
             CLEAN_WORKSPACE: cleanWorkspace,
             CLEAN_WORKSPACE_AFTER: cleanWsAfter,
             CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY_AFTER: cleanWorkspaceBuildOutputAfterBuild
--- a/pipelines/build/common/config_regeneration.groovy
+++ b/pipelines/build/common/config_regeneration.groovy
@@ -443,6 +443,7 @@ class Regeneration implements Serializable {
                 ENABLE_TESTDYNAMICPARALLEL: DEFAULTS_JSON['testDetails']['enableTestDynamicParallel'] as Boolean,
                 ENABLE_INSTALLERS: true,
                 ENABLE_SIGNER: true,
+                ENABLE_SBOM: true,
                 CLEAN_WORKSPACE: true,
                 CLEAN_WORKSPACE_AFTER: true,
                 CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY_AFTER: false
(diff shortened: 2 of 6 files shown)


sophia-guo:
I'm not sure if we should add this new parameter. @andrew-m-leonard my understanding is --create-sbom is a new BUILD_ARGS, which can be used by makejdk-any-platform.sh to generate SBOM information( either console output or json file). The default is 'false'. 

We already have parameter [String additionalBuildArgs](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/build_base_file.groovy#L54), which could be used to set [extra BUILD_ARGS](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/build_base_file.groovy#L116-L118) for each run. If we'd like to turn it on for some specific platform or version we could also just update the groovy config file buildArgs.

sophia-guo:
The default is set as fault https://github.com/adoptium/temurin-build/blob/master/sbin/common/config_init.sh#L471. 

For pipeline job to enable it we can use parameter **additionalBuildArgs** https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk17-pipeline/build?delay=0sec

For build job we can use  **BUILD_ARGS** https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-x64-temurin/build?delay=0sec

Only if we'd like to set it as true by default we can update configuration files https://github.com/adoptium/ci-jenkins-pipelines/tree/master/pipelines/jobs/configurations by **buildArgs**.

zdtsw:
> The default is set as fault https://github.com/adoptium/temurin-build/blob/master/sbin/common/config_init.sh#L471.
> 
> For pipeline job to enable it we can use parameter **additionalBuildArgs** https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk17-pipeline/build?delay=0sec
> 
> For build job we can use **BUILD_ARGS** https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-x64-temurin/build?delay=0sec
> 
> Only if we'd like to set it as true by default per version or platform we can update configuration files https://github.com/adoptium/ci-jenkins-pipelines/tree/master/pipelines/jobs/configurations by **buildArgs**.
> 
> If we'd like to set all as true by default we can update https://github.com/adoptium/temurin-build/blob/master/sbin/common/config_init.sh#L471. Though I don't think it's necessary for now.

Yeah, I agree! 
To make it flexible, I did the last change in jdk8/11/17/18/19 config groovy files, which is to enable all temurin build with this flag as part of the build arg.
and for each run, if we want to re-run or do a test run, can set it from UI.

sophia-guo:
> If we'd like to set all as true by default we can update https://github.com/adoptium/temurin-build/blob/master/sbin/common/config_init.sh#L471. Though I don't think it's necessary for now.

I mean if we need to update all jdk8/11/17/18/19 config groovy files we can directly update [the default to be true](https://github.com/adoptium/temurin-build/blob/master/sbin/common/config_init.sh#L471).

zdtsw:
@andrew-m-leonard Do you see we should have all JDK versions * arch to enable --create-sbom? 
If that is the case, to change default value to "[true](https://github.com/adoptium/temurin-build/blob/master/sbin/common/config_init.sh#L471)" is the easiest way to go.   
Otherwise, we  add --create-sbom to all buildArgs(as-is in this PR) to go forward. 
Even they archive the same purpose

smlambert:
I think the default should be true and if there are cases that we do not want it, then we would change the buildArgs to be false on a case by case basis.  I know of no cases where we'd want to set it to false at this moment.

zdtsw:
> I think the default should be true and if there are cases that we do not want it, then we would change the buildArgs to be false on a case by case basis. I know of no cases where we'd want to set it to false at this moment.

alright, then let me do the simple solution to change the default value for the time being.

zdtsw:
new PR : https://github.com/adoptium/temurin-build/pull/2926

PR #297, opened 2022-05-16
feat: enable --create-sbom (in temurin-build) by passing new flag into BUILD_ARGS

zdtsw:
- enable all OS except Windows(under test)

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -42,6 +42,7 @@ class Builder implements Serializable {
     boolean enableTestDynamicParallel
     boolean enableInstallers
     boolean enableSigner
+    boolean enableSBOM
     boolean publish
     boolean release
     String releaseType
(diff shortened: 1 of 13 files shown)


karianna (changes requested):
My understanding is that SBOM only works for Java 19+. Or is that just the reproducible builds feature of that?

andrew-m-leonard:
@zdtsw So there is no reason the SBOM feature can't be used for all the versions, although it's intended purpose is to work along side reproducible builds. Reproducible builds is currently jdk-19, and also jdk-17 in the near future (possibly 17.0.4).

From a pure SSDF point of view we want to use it for all versions.

andrew-m-leonard (approved):

andrew-m-leonard:
> My understanding is that SBOM only works for Java 19+. Or is that just the reproducible builds feature of that?

@karianna I think you maybe thinking of the build-script changes to enable reproducible builds that require jdk-19+ due to requirement for the bootjdk jar -date option which only exists in jdk-19 and jdk-17.0.3. SBOM support is actually independent of this.

karianna (approved):

PR #305, opened 2022-05-19
fix: wrong flag

zdtsw:
- missing "-" for "--create-jre-image" on jdk19

### DIFF
--- a/pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
@@ -90,9 +90,6 @@ class Config17 {
                 os                  : 'windows',
                 arch                : 'x86-32',
                 additionalNodeLabels: 'win2012&&vs2017',
-                buildArgs           : [
-                        temurin : '--jvm-variant client,server'
-                ],
                 test                : 'default',
                 buildArgs           : [
                         "temurin"   : '--create-jre-image --create-sbom'
--- a/pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
@@ -92,7 +92,7 @@ class Config17 {
                 additionalNodeLabels: 'win2012&&vs2017',
                 test                : 'default',
                 buildArgs           : [
-                        "temurin"   : '--create-jre-image --create-sbom'
+                        "temurin"   : '--jvm-variant client,server --create-jre-image --create-sbom'
                 ]
         ],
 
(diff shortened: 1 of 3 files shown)


zdtsw:
@andrew-m-leonard @sxa  Can I get a quick review on this?

pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
```diff
@@ -90,9 +90,6 @@ class Config17 {
                 os                  : 'windows',
                 arch                : 'x86-32',
                 additionalNodeLabels: 'win2012&&vs2017',
-                buildArgs           : [
-                        temurin : '--jvm-variant client,server'
```
karianna, line 94: I think the 17+ build arg is `--with-jvm-variants` so could replace with that as I'm not sure we want to remove the intent here?

karianna (changes requested):

pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
sxa: @karianna This is buildArgs rather than a configureArg, and we do list `--jvm-variant` on makejdk-any-platform.sh which presumably (hopefully...) maps down to the option you're referring to

zdtsw: yeah, I am not sure if "--jvm-variant" should be "--with-jvm-variants" but the current config is not using either. 
@andrew-m-leonard could you confirm if we need the flag on 17+ ?

andrew-m-leonard: @zdtsw this is BUILD_ARGS which gets processed here: https://github.com/adoptium/temurin-build/blob/6a642bc6aa9f5b4530745e831c35d5c174c51a13/sbin/common/config_init.sh#L346
So above looks correct

sxa: @karianna IIRC it was normal for us to include the `client` variant on the 32-bit platforms. Obviously happy for that to be revisited if you feel it's necessary though.

sxa: @karianna Client VM for Win32 was originally requested and added via https://github.com/adoptium/temurin-build/issues/871
Was added for Win64 in https://github.com/adoptium/temurin-build/pull/1113 and reverted in https://github.com/adoptium/temurin-build/pull/2062

karianna: Right, yes we *do* need it for Win32 binaries else those legacy apps won't run.  But we *do not* need it for x64

karianna (approved):
LGTM

PR #306, opened 2022-05-19
Add support for Jenkins built-in node

vsebe:
Add built-in label to the pipelines due to the renaming of "master node"
to "built-in node" in newer Jenkins versions (master label to be removed
after Jenkins upgrade).

Signed-off-by: Violeta Sebe <vsebe@ca.ibm.com>

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -771,8 +771,8 @@ class Builder implements Serializable {
 
                             if (downstreamJob.getResult() == 'SUCCESS') {
                                 // copy artifacts from build
-                                context.println "[NODE SHIFT] MOVING INTO MASTER NODE..."
-                                context.node("master") {
+                                context.println "[NODE SHIFT] MOVING INTO CONTROLLER NODE..."
+                                context.node("built-in || master") {
                                     context.catchError {
 
                                         //Remove the previous artifacts
@@ -813,7 +813,7 @@ class Builder implements Serializable {
 
                                     }
                                 }
-                                context.println "[NODE SHIFT] OUT OF MASTER NODE!"
+                                context.println "[NODE SHIFT] OUT OF CONTROLLER NODE!"
                             } else if (propagateFailures) {
                                 context.error("Build failed due to downstream failure of ${downstreamJobName}")
                                 currentBuild.result = "FAILURE"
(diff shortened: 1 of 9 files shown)


vsebe:
@AdamBrousseau please review, thank you!

pipelines/build/common/build_base_file.groovy
```diff
@@ -771,8 +771,8 @@ class Builder implements Serializable {
 
                             if (downstreamJob.getResult() == 'SUCCESS') {
                                 // copy artifacts from build
-                                context.println "[NODE SHIFT] MOVING INTO MASTER NODE..."
-                                context.node("master") {
+                                context.println "[NODE SHIFT] MOVING INTO BUILT-IN NODE..."
```
AdamBrousseau, line 774: It think this one should be `Controller` to match the other uses.

AdamBrousseau (changes requested):
We should add a note either in the PR description or commit comment indicating that we will come along and remove `master` in the future once everyone using the scripts has signed off that they have updated their Jenkins instances.
Other than that the change looks good to me assuming we've caught all the uses.

sxa:
I'm ok with this for now, but I will note that I plan to try and shift away from running anything on this node.
@vsebe Was there something not working as a result of the change that this PR fixes or was it just to be consistent with the new naming? I haven't seen a backlog of jobs waiting for `master` (Probably because the build-in node still identifies as such in https://ci.adoptopenjdk.net/label/master/)

AdamBrousseau:
More info:
https://www.jenkins.io/doc/book/managing/built-in-node-migration/

Your Manage Jenkins page should have a paragraph about applying the change. Basically the label of the Controller will change once you `Apply Migration`. I don't believe there is a Jenkins due date to migrate but we have an IBM due date. I agree that running anything on the controller is a separate issue. We've had instances where it was strictly not allowed but we also have some infra type jobs that run things on it for specific reasons.

pipelines/build/common/build_base_file.groovy
vsebe: Updated.

karianna:
We'll need to time this at Eclipse Adoptium.  Stewart and I were planning on trying the migration on Monday.  We should probably land this first, see if it all blends, ask the community if anyone else is running internal Jenkins, and then remove the old master wording.

karianna (approved):
LGTM

sxa (commented):
Is `context.node('built-in || master') {` definitely ok? I would normally have written that without spaces in the condition and it may look a bit strange to people to see it like this.

AdamBrousseau:
Spaces are normally fine. I'm not sure if Violeta tested this specific piece of code but we use spaces in existing code.

vsebe:
Spaces are fine in node labels conditions. Yes, I tested it internally.

AdamBrousseau (approved):

AdamBrousseau:
@sxa any more concerns or can we get this merged?

sxa (approved):
LGTM. Now that we're into a new week I'm happy for this to go in

PR #317, opened 2022-06-01
Clean workspace prior to installer jobs

andrew-m-leonard:
The current buildInstallerMac/Windows() stages relied on the workspace being clean by the prior sign_build stage running on the same Jenkins master node. Now with 2 instances, it may not run on the same node. Thus this PR ensures each stage performs the workspace clean.

Fixes: https://github.com/adoptium/temurin-build/issues/2959

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -636,8 +636,11 @@ class Build {
 
         context.node('built-in || master') {
             context.stage("installer") {
+                // Ensure master context workspace is clean
+                context.sh "rm workspace/target/* || true"
+
                 switch (buildConfig.TARGET_OS) {
-                    case "mac": context.sh 'rm -f workspace/target/*.pkg workspace/target/*.pkg.json workspace/target/*.pkg.sha256.txt'; buildMacInstaller(versionData); break
+                    case "mac":     buildMacInstaller(versionData); break
                     case "windows": buildWindowsInstaller(versionData); break
                     default: break
                 }
@@ -665,6 +668,9 @@ class Build {
 
         context.node('built-in || master') {
             context.stage("sign installer") {
+                // Ensure master context workspace is clean
+                context.sh "rm workspace/target/* || true"
+
                 if (buildConfig.TARGET_OS == "mac" || buildConfig.TARGET_OS == "windows") {
                     try {
                         signInstallerJob(versionData);
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -636,8 +636,8 @@ class Build {
 
         context.node('built-in || master') {
             context.stage("installer") {
-                // Ensure master context workspace is clean
-                context.sh "rm workspace/target/* || true"
+                // Ensure master context workspace is clean of any previous archives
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
Test job: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk-mac-x64-temurin/4/

andrew-m-leonard:
run tests

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -636,8 +636,11 @@ class Build {
 
         context.node('built-in || master') {
             context.stage("installer") {
+                // Ensure master context workspace is clean
+                context.sh "rm workspace/target/* || true"
```
karianna, line 640: Should that be rm -rf

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -665,6 +668,9 @@ class Build {
 
         context.node('built-in || master') {
             context.stage("sign installer") {
+                // Ensure master context workspace is clean
+                context.sh "rm workspace/target/* || true"
```
karianna, line 672: Should this be rm -rf?

karianna (changes requested):

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: interesting, I actually copied that from further up... trying to think why it's done this way..?

andrew-m-leonard: intention is just to clear any previous archives, so -r not required, always preferable to not to use -r
-f not really necessary as there is not terminal on a Jenkins job, but I think I will add it to be correct

andrew-m-leonard:
run tests

karianna (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -665,6 +668,9 @@ class Build {
 
         context.node('built-in || master') {
             context.stage("sign installer") {
+                // Ensure master context workspace is clean of any previous archives
+                context.sh "rm -f workspace/target/* || true"
```
sxa, line 672: `|| true` is pointless when running with `-f` so I would suggest removing it
```suggestion
                context.sh "rm -f workspace/target/*"
```

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -636,8 +636,11 @@ class Build {
 
         context.node('built-in || master') {
             context.stage("installer") {
+                // Ensure master context workspace is clean of any previous archives
+                context.sh "rm -f workspace/target/* || true"
```
sxa, line 640:
```suggestion
                context.sh "rm -f workspace/target/*"
```

sxa (approved):
I'm ok to approve this but would prefer if the `|| true` was removed from the `rm -f` commands.

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: @sxa i'm not sure it is, if there are no files then the exit code will be set to "1" on the rm -f, causing stage failure ?

PR #320, opened 2022-06-08
Implement GPG signing of the binary build artifacts

sxa:
First stage in solving https://github.com/adoptium/temurin-build/issues/1275

Will require:
- a change in [refactor_openjdk_release_tool](https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/refactor_openjdk_release_tool/) for them to be published ([Done](https://github.com/adoptium/github-release-scripts/pull/88))
- Changes in the API/website to make them downloadable ([Done](https://github.com/adoptium/api.adoptium.net/pull/464))
- Likely subsequent changes in the underlying signing job which should be moved into an external script
- Potential changes to where the job is run to avoid deadlocks occurring with the `master` nodes in jenkins (Hasn't been a problem ... so far!)
- Documentation on how users can use these signatures once they are made available.
- Potentially look at changing this to run from the individual platform build pipelines instead of the top level one - currently the sha256sums are only generated in the top level pipeline so are not available before that point (Except on mac/windows - [slack thread)](https://adoptium.slack.com/archives/C09NW3L2J/p1654609744697389) (Done in this PR)
- Look at aligning more with how the OpenJ9 team have done this as they cannot currently reuse the contents of this PR and will be skipping it

But this is the first step and is a prerequisite for the others. Note that this functionality will always be executed and is not currently gated by the `enableSigner` checkbox on the pipeline jobs. That can be added in a subsequent PR if desired.

This has been successfully trialled on the pipeline jobs.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -496,7 +496,7 @@ class Build {
                             selector: context.specific("${signJob.getNumber()}"),
                             filter: 'workspace/target/*',
                             fingerprintArtifacts: true,
-                            target: "workspace/target/",
+                            target: 'workspace/target/',
                             flatten: true)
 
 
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM - do we need to keep this in draft so we execute the right chain of events?

sxa:
> LGTM - do we need to keep this in draft so we execute the right chain of events?

No, this can safely go in without the other bits being present. It just won't put the artifacts anywhere other than in jenkins at the moment.

tellison:
Sign the hashcode files - that's unusual isn't it?
I'd only expect a signature for the release artefacts themselves.

sxa:
> Sign the hashcode files - that's unusual isn't it? I'd only expect a signature for the release artefacts themselves.

Not really - here are a couple of examples where the has is what's signed:
- https://github.com/nodejs/node#verifying-binaries
- https://wiki.eclipse.org/Platform-releng/How_to_check_integrity_of_downloads

The GPG verification process will be more iCPU ntensive than the SHA sum (and our files are quite big, but I appreciate that on modern hardware it likely isn't too much of an overhead) and this way you're getting some guarantee that both are integral without having to clog up the number of artifacts we have available to download by having double the number of GPG signature files.

Obviously either can be technically done (I initially prototyped doing this with the full archive file, but switched over after a recommendation from Eclipse)

sxa:
@tellison FYI The artifacts that get signed aren't affected by this PR. I've adjusted the jenkins job that does the work to pick up all the tar.gz/zip/msi/pkg files instead of the SHAsums and this PR does not require modification to pick up those instead of the SHAs

tellison (approved):

sxa:
@AdamBrousseau Does OpenJ9 use the signing options in these pipelines and would therefore be affected by this unless an additional guard was put in place around it?

sxa:
> @sxa will let you merge at a good time to monitor it.

Yeah I'm holding off pending verification that it can be done without breaking OpenJ9

Issue #328, opened 2022-06-16
sbom cannot be query directly from api.adoptium.net


zdtsw:
**Expected behaviour:**
when using api.adoptium.net to query for sbom, should get correct return from swagger and using curl/wget

**Observed behaviour:**
on /binary endpoint
```
curl -iL https://api.adoptium.net/v3/binary/latest/17/ea/linux/x64/sbom/hotspot/normal/eclipse?project=jdk
HTTP/1.1 404 Not Found
```
on /assets endpoint
`https://api.adoptium.net/v3/assets/latest/17/hotspot?architecture=x64&image_type=sbom&os=linux&vendor=eclipse`
server response []
client response with all other binary_type(e.g jdk jre debugimage testimage etc) but not "sbom"

**Any other comments:**
The root cause probably is not in the api.adoptium.net but the metadata it uses to map from /api/v3/assets or /api/v3/binary is missing.
Suspect this is caused when we made the change from old sbom file format from tar.gz to json, which broke the [function](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L756) 
so in the case of sbom file (e.g OpenJDK17U-sbom_x64_linux_hotspot_2022-06-10-09-44.jsons) it wont generate a metadata json file

zdtsw:
now we have two json files generated
![Screenshot from 2022-06-19 07-41-13](https://user-images.githubusercontent.com/915053/174467556-f01bdbe5-3ae7-4ec9-ae27-1a7741ec97f2.png)

zdtsw:
![Screenshot from 2022-06-21 07-33-49](https://user-images.githubusercontent.com/915053/174723640-ef714db0-3f82-49a0-9381-95e214d2c04f.png)
now release binary git repo has two json files.

zdtsw:
issue has been resolved, finally updated pushed to production api server and everything works fine.

PR #329, opened 2022-06-16
feat: rename metadata file

zdtsw:
- previous named as *sbom<XXX>.json
	- now change to *sbom<XXX>-metadata.json
	- update filter on binary files to include json (sbom<XXX>.json
	- this could solve problem with *sbom*.json file which does not generate metadata any more
	- this could help "api" to filter sbom's  metadata
Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/328

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1051,10 +1051,12 @@ class Build {
             }
 
             // Special handling for sbom metadat file (to be backwards compatible for api service)
+            // from "*sbom<XXX>.json" to "*sbom<XXX>-metadata.json"
             if (file.contains("sbom")) {
-                file.replace(".json", "-metadata") // from "*sbom<XXX>.json" to "*sbom<XXX>-metadata"
+                context.writeFile file: file.replace(".json", "-metadata.json"), text: JsonOutput.prettyPrint(JsonOutput.toJson(data.asMap()))
+            } else {
+                context.writeFile file: "${file}.json", text: JsonOutput.prettyPrint(JsonOutput.toJson(data.asMap()))
             }
-            context.writeFile file: "${file}.json", text: JsonOutput.prettyPrint(JsonOutput.toJson(data.asMap()))
         })
     }
 
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1050,7 +1050,7 @@ class Build {
                 metaWrittenOut = true
             }
 
-            // Special handling for sbom metadat file (to be backwards compatible for api service)
+            // Special handling for sbom metadata file (to be backwards compatible for api service)
             // from "*sbom<XXX>.json" to "*sbom<XXX>-metadata.json"
             if (file.contains("sbom")) {
                 context.writeFile file: file.replace(".json", "-metadata.json"), text: JsonOutput.prettyPrint(JsonOutput.toJson(data.asMap()))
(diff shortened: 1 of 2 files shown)


zdtsw:
How you feel about this PR? @smlambert @andrew-m-leonard 
Could rename break anything else? 
some discussion https://adoptium.slack.com/archives/CFUJV9XV0/p1655367001333359?thread_ts=1653911822.234059&cid=CFUJV9XV0

andrew-m-leonard:
@zdtsw the sbom and the old metadata file are two different things, are we sure we are not confusing things here?

zdtsw:
> @zdtsw the sbom and the old metadata file are two different things, are we sure we are not confusing things here?

so I think in the past we only create *.tar.gz or *.zip/pkg etc file, and then based on these filename, we create *.json (which is the metadata)
then we added support for sbom, which is named as sbom*.json, so the function does not create metadata for the sbom file.
so the api part is not getting sbom's metadata file.

what I am thinking is to, keep sbom as the sbom*.json file, but make all the metadata json file from "*.json" to  "*-metadata.json"
So in case of sbom, it should have a sbom*.json and its metadata sbom*-metadata.json

Then in API, we only filter out *-metadata.json

andrew-m-leonard:
It would be good to see the User scenario that has a problem, for example API query to get SBOM is failing: ....exception...?

zdtsw:
> It would be good to see the User scenario that has a problem, for example API query to get SBOM is failing: ....exception...?

The current problem is, we do not generate metadata file for sbom. 
that is causing api service cannot download sbom json nor get assets of it (because of missing metadata on sbom)


e.g https://api.adoptium.net/v3/binary/latest/18/ga/linux/x64/sbom/hotspot/normal/eclipse?project=jdk
return 404 Error: Not Found
e.g https://api.adoptium.net/v3/assets/latest/18/hotspot?architecture=x64&image_type=sbom&os=linux&vendor=eclipse
return 200 with [] body

andrew-m-leonard:
@zdtsw ok lets discuss as I don't really want a second json file for the sbom.json, seems unnecessary, lets try and fix the API server so it doesn't need it... surely it can just serve up the sbom.json....?
Seems a bit silly to effectively have artifacts for:
```
sbom.json
sbom-metadata.json
```

zdtsw:
If I no mistaken, we need to add all these fields into sbom.json in order for it to be severed as a metadata 
https://github.com/adoptium/api.adoptium.net/blob/main/adoptium-updater-parent/adoptium-datasources-parent/adoptium-github-datasource/src/main/kotlin/net/adoptium/api/v3/dataSources/github/graphql/models/GHMetaData.kt#L30
(warning is not a mandatory filed)

andrew-m-leonard:
Ah right, starting to understand the problem, let me have a think

zdtsw:
After discussion:
only add -metadata into sbom's metadata file.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1050,6 +1050,10 @@ class Build {
                 metaWrittenOut = true
             }
 
+            // Special handling for sbom metadat file (to be backwards compatible for api service)
+            if (file.contains("sbom")) {
+                file.replace(".json", "-metadata") // from "*sbom<XXX>.json" to "*sbom<XXX>-metadata"
```
andrew-m-leonard, line 1055: not sure you want to inline replace file variable, I would put the context.writeFile line in both sides of an if (file.contains("sbom")) {context.writeFile file:file.replace(".json", "-metadata") .....} else {..as before.}

andrew-m-leonard (approved):
looks good

zdtsw:
@smlambert if you have time, could you do a review on it and if good to merge it? want to have it tested out and see if it works with api, thanks!

smlambert (approved):
thanks @zdtsw !

PR #331, opened 2022-06-17
Ensure archive filename OpenJDKU prefix ends with U

andrew-m-leonard:
Branched new jdk versions eg.jdk19, do not have a "U" suffix on the java to build archive filename.
For compatibility we need it to end with U: OpenJDK19U

Fixes: https://github.com/adoptium/temurin-build/issues/2980

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1074,7 +1074,7 @@ class Build {
 
         javaToBuild = javaToBuild.trim().toUpperCase()
  
-        # Add "U" to javaToBuild filename prefix for non-head versions
+        // Add "U" to javaToBuild filename prefix for non-head versions
         if (!javaToBuild.endsWith("U") && !javaToBuild.equals("JDK")) {
           javaToBuild += "U"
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
Test build: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk19-linux-x64-temurin/2/

karianna (approved):
LGTM

karianna (commented):
Hmm, I'm not sure this is a desired change?  Does this change the file name for our end users from what they are used to?

gdams (approved):

sxa:
> Hmm, I'm not sure this is a desired change? Does this change the file name for our end users from what they are used to?

For the last cycle we tried to build from the `u` repository for the first build of a new version and that caused us problems. Because of that, we're going to build from the non-u repository for the first release, and allow the `U` to be added to the resulting filename in order to ensure more consistency in the filename for the end user but making the pattern always `OpenJDKXXU`. This is consistent with the [jdk18+36](https://github.com/adoptium/temurin18-binaries/releases/tag/jdk-18%2B36) release.

Issue #234, opened 2022-01-25
Adoptium top level pipeline build should reflect all child job results


llxia:
Currently, Adoptium top-level pipeline build (i.e., openjdk8-pipeline, openjdk11-pipeline, etc) does not reflect all child job results, especially not including AQA test build results. It causes confusing/misleading information. For example:

- If the top-level pipeline build is SUCCESS/green, it does not mean all AQA tests jobs are passed.
- If the top-level pipeline build is completed, it does not mean all AQA test jobs are completed. 

Adoptium top-level pipeline build should reflect all child job results so that the triage work can be easier. For example, if any AQA test build fails, the top-level pipeline build status should be FAILURE or UNSTABLE.

smlambert:
related: https://github.com/adoptium/ci-jenkins-pipelines/issues/11

karianna:
FYI - I think we'll need to provide the upgrade Jenkins to resolve this.

smlambert:
As far as I am aware, this issue can be resolved by updating pipeline code and does not require a Jenkins upgrade to accomplish it.

llxia:
Does anyone know why we have `propagate: false` for the test jobs?
https://github.com/adoptium/ci-jenkins-pipelines/blob/cee0b690b2a3a199b1f3f77892b6a1e18a20ea3b/pipelines/build/common/openjdk_build_pipeline.groovy#L292

smlambert:
The only reason was historically, we did not have many clean runs.  Now that we have cleaned up the test jobs (a lot), we should be switching that to true.  That may be sufficient to resolve this issue (to have top-level pipeline build reflect test job status at least).  Other child jobs like smoke tests and installers would need their own adjustments.

andrew-m-leonard:
I think the following would need updating:
- Smoketest propagate set to true
- AqaTests popagate set to true
- The pipeline templates need propagateFailures defaulting to true
   - https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/jobs/pipeline_job_template.groovy
   - https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/jobs/release_pipeline_job_template.groovy

smlambert:
Related: #121 #195

smlambert:
@chandrashritii - can you add a comment to this issue, so we can add you as an assignee?

chandrashritii:
[...]

PR #333, opened 2022-06-20
Set propagate to true for AQA test builds

llxia:
resolves: https://github.com/adoptium/ci-jenkins-pipelines/issues/234

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -406,7 +406,7 @@ class Build {
                         }
                         context.catchError {
                             context.build job: jobName,
-                                    propagate: false,
+                                    propagate: true,
                                     parameters: [
                                             context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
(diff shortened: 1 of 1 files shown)


llxia:
@smlambert @sophia-guo could you review it? Thanks

karianna (approved):
LGTM

sophia-guo:
Set AQA test builds `propagate to true` can propagate AQA tests result to build jobs, i.e., jdk**-**_temurin job), which will not propagate result to Adoptium top-level pipeline build. https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/build_base_file.groovy#L772.

smlambert (changes requested):
Before this gets merged, I would like to talk about the side-effects.

smlambert:
In discussion with @sophia-guo , wondering about whether this then means:
- that the parent will not archive the build artifacts
- if we change the pipeline status (not just build job status) whether it means that other child jobs will not get run

llxia:
Is there a way for us to run a test build to confirm at Adoptium Jenkins?

llxia:
Close this PR until we find a better solution for this.

PR #310, opened 2022-05-20
chore: Support for tag in source repo when it is tagged as _adoptium

zdtsw:
We currently do "release" build and remove  ending "_adopt" from source code tag, e.g: https://github.com/adoptium/jdk18u/tags

Should consider to tag it (when merge from master branch to release branch) as "_adoptium" in the future. 
Thus, this PR is to add support for both.

PR #334, opened 2022-06-21
Revert "chore: Support for tag in source repo when it is tagged as _a…

zdtsw:
…doptium (#310)"

This reverts commit f236cbd86848b3240af7938220518e7e8f54e5f4.

along with https://github.com/adoptium/mirror-scripts/pull/19


zdtsw:
@andrew-m-leonard think we need this one too

andrew-m-leonard (approved):

PR #335, opened 2022-06-21
fix Visual Studio version labels for Windows

gdams:
We are already building JDK17+ on VS 2019 so the label was incorrect.

Also updated JDK11 so this PR can be merged once https://github.com/adoptium/temurin-build/pull/2992 has been merged

### DIFF
--- a/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
@@ -56,7 +56,7 @@ class Config11 {
                 os                  : 'windows',
                 arch                : 'x64',
                 additionalNodeLabels: [
-                        temurin:    'win2012&&vs2017',
+                        temurin:    'win2012&&vs2019',
                         openj9:     'win2012&&vs2017',
                         dragonwell: 'win2012'
                 ],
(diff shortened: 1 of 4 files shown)


sxa (commented):
Couple of points:

1. In general the playbooks should cause all compliler versions to be on any suitable build machine, so that conditional shouldn't be necessary.
2. At the moment there are only the [two azure systems tagged as vs2019](https://ci.adoptopenjdk.net/label/win2012&&vs2019/) which is going to limit the range of which machines it can build on which is presumably not necessary. Can we check that the other machines are ok for building on 2019 to ensure we have multiple providers capable of doing these builds?

gdams:
@sxa I've added the VS2019 label to the IBM machines (they both have it installed so we should be okay now)

sxa (approved):
> @sxa I've added the VS2019 label to the IBM machines (they both have it installed so we should be okay now)

👍🏻 I would imagine thuat https://ci.adoptopenjdk.net/computer/build%2Dalibaba%2Dwin2012r2%2Dx64%2D1/ could likely have the label added too, but on the basis that we have two providers labelled I'm fine with approving this

karianna (approved):
LGTM

PR #338, opened 2022-06-23
Add jdk20 pipeline config

andrew-m-leonard:
Based from jdk19 config.

### DIFF
--- a/pipelines/jobs/configurations/jdk20.groovy
+++ b/pipelines/jobs/configurations/jdk20.groovy
@@ -1,0 +1,59 @@
+targetConfigurations = [
+        "x64Mac"      : [
+                "temurin"
+        ],
+        "x64Linux"    : [
+                "temurin"
+        ],
+        "x64AlpineLinux" : [
+                "temurin"
+        ],
+        "x64Windows"  : [
+                "temurin"
+        ],
+        "x32Windows"  : [
+                "temurin"
+        ],
+        "ppc64Aix"    : [
+                "temurin"
+        ],
+        "ppc64leLinux": [
+                "temurin"
+        ],
+        "s390xLinux"  : [
+                "temurin"
+        ],
+        "aarch64Linux": [
+                "hotspot",
+                "temurin"
+        ],
+        "aarch64AlpineLinux" : [
+                "temurin"
+        ],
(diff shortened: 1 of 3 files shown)


pipelines/jobs/configurations/jdk20.groovy
```diff
@@ -0,0 +1,59 @@
...
+                "temurin"
+        ],
+        "ppc64leLinux": [
+                "temurin"
+        ],
+        "s390xLinux"  : [
+                "temurin"
+        ],
+        "aarch64Linux": [
+                "hotspot",
```
gdams, line 27: what's the reason for the `hotspot` build?

andrew-m-leonard: it's from jdk-19, added here: https://github.com/adoptium/ci-jenkins-pipelines/pull/251
@sxa presume this is actually good here since this is jdk(head), but should we remove it from jdk-19 ?

sxa: Yes - I'd say so. Not sure if we can add inline comments into these files but might be good to put a note saying "I'f you've moved this out of JDK head you might want to remove this one"

pipelines/jobs/configurations/jdk20_pipeline_config.groovy
```diff
@@ -0,0 +1,185 @@
+class Config20 {
+  final Map<String, Map<String, ?>> buildConfigurations = [
+        x64Mac    : [
+                os                  : 'mac',
+                arch                : 'x64',
+                additionalNodeLabels: 'macos10.14',
+                additionalTestLabels: [
+                        openj9      : '!sw.os.osx.10_11'
```
gdams, line 8: It would be nice to rip out the OpenJ9 specific sections from the configurations here if possible but lets merge for now. Maybe we can do this in a follow-up PR?

gdams (approved):

pipelines/jobs/configurations/jdk20_pipeline_config.groovy
smlambert: I believe these configurations are still in use, so I would not want to see them removed as it would break some things.

PR #346, opened 2022-06-29
feat: add build env container's image SHA1

zdtsw:


karianna (approved):
LGTM

gdams:
Will this only be used when we are using a docker image? The logic seems unclear to me?

zdtsw:
> Will this only be used when we are using a docker image? The logic seems unclear to me?

it should happen for all cases: build inside container or not.

docker.txt either get a SHA1 or "" (default value from string, size 1byte)
[PR](https://github.com/adoptium/temurin-build/pull/3005) in temurin-build, should read from docker.txt , either show SHA! or with "N.A" in sbom.

andrew-m-leonard (approved):

PR #358, opened 2022-07-06
feat: add comparision for macos

zdtsw:
source from :https://github.com/adoptium/temurin-build/issues/2899#issuecomment-1149963539

Ref: https://github.com/adoptium/temurin-build/issues/3015


tools/reproduce_comparison/compareMacOS.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+set -e
+
+help()
```
karianna, line 5: I think there's an std shell/unix command line pattern to use in help to indicate mandatory or optional parameters, e.g.:

`usage: cp [-R [-H | -L | -P]] [-fi | -n] [-aclpsvXx] source_file target_file`

tools/reproduce_comparison/compareMacOS.sh
```diff
@@ -0,0 +1,103 @@
...
+   echo "Syntax: compareMacOS.sh"
+   echo "arguments:"
+   echo "$1     Directory to one JDK"
+   echo "$2     Directory to another JDK"
+   echo "$3     Path to local certificate file"
+   echo "$4     Skip final compare step, used for debug or from Jenkins. default to run comparision"
+}
+
+if [ ! -f "$3" ]; then
+    echo "Panic: Require valid local Certificate specified"
```
karianna, line 18: Not sure if the use of 'Panic:' is the wording we want to use here :-).

tools/reproduce_comparison/compareMacOS.sh
```diff
@@ -0,0 +1,103 @@
...
+
+    if [ ! -d "${JDK_DIR}/Contents" ]; then
+        echo "Panic: $JDK_DIR does not contain the MacOS JDK Contents directory"
+        exit 1
+    fi
+
+    FILES=$(find "${JDK_DIR}" \( -type f -and -path '*.dylib' -or -path '*/bin/*' -or -path '*/lib/*' -not -path '*/modules_expanded/*' -or -path '*/jpackageapplauncher*' \))
+    for f in $FILES
+    do
+        echo "# Removing any signature from $f"
```
karianna, line 83:
```suggestion
        echo "# Removing any signatures from $f so we can compare the JDKs"
```

tools/reproduce_comparison/compareMacOS.sh
```diff
@@ -0,0 +1,103 @@
...
+        codesign --remove-signature $f
+        echo "# Signing $f with local certificate"
```
karianna, line 85:
```suggestion
        echo "# Signing $f with a local certificate so we can compare the JDKs"
```

tools/reproduce_comparison/compareMacOS.sh
```diff
@@ -0,0 +1,103 @@
...
+        # Sign both with same local Certificate, this adjusts __LINKEDIT vmsize identically
+        codesign -s "$CERT" --options runtime -f --timestamp "$f"
+        echo " # Remove local Certificate from $f"
+        codesign --remove-signature $f
+    done
+
+    echo "Done Removing Signature!"
+done
+echo "All preparion done!"
```
karianna, line 94:
```suggestion
echo "All preparation done!"
```

tools/reproduce_comparison/compareMacOS.sh
```diff
@@ -0,0 +1,103 @@
...
+
+    FILES=$(find "${JDK_DIR}" \( -type f -and -path '*.dylib' -or -path '*/bin/*' -or -path '*/lib/*' -not -path '*/modules_expanded/*' -or -path '*/jpackageapplauncher*' \))
+    for f in $FILES
+    do
+        echo "# Removing any signature from $f"
+        codesign --remove-signature $f
+        echo "# Signing $f with local certificate"
+        # Sign both with same local Certificate, this adjusts __LINKEDIT vmsize identically
+        codesign -s "$CERT" --options runtime -f --timestamp "$f"
+        echo " # Remove local Certificate from $f"
```
karianna, line 88:
```suggestion
        echo " # Remove local Certificate from $f so we can compare JDKs easier."
```

tools/reproduce_comparison/compareMacOS.sh
```diff
@@ -0,0 +1,103 @@
...
+echo "All preparion done!"
+
+if [ "$4" == "skip" ]; then
+    echo "Skip final comparision"
+    exit 0
+fi
+
+echo "Start Comparing JDKs $1 and $2"
+diff -q -r "$1" "$2"
+echo "Done Comparing: JDKs are identical!"
```
karianna, line 103: Don't we need to read the result of the diff before claiming this?

karianna (changes requested):

tools/reproduce_comparison/compareMacOS.sh
zdtsw: since we set "set -e"
if $1 and $2 aint the same, it prints "**** and *** differ" and exit non-0
otherwise, prints "Done Comparing: JDKs are identical!" and exit 0

karianna: @zdtsw got it, but I think that's slightly brittle as `set -u` could easily be removed by someone later on and this woyuld silently 'always succeed'

zdtsw: updated with exit code.
even "set" is moved,  should have exit code to determinate if script successfully diff or not.

difference is 
with "set -e" in a positive case, prints out `Done Comparing: Files X and Y differ`
in a negative case, print out nothing after `Start Comparing JDKs X and Y`

andrew-m-leonard (approved):
Looks good, thanks @zdtsw

PR #360, opened 2022-07-07
Ensure node context workspace is clean prior to gpg signing

andrew-m-leonard:
Fixes: https://github.com/adoptium/temurin-build/issues/3029

The problem of the previous build .sig artifact is because the gpgSign() logic in the build job switches to the gpgsign node context, and then copys the new sign_temurin_gpg output but without clearing any of the previous build output from the workspace/target context for the build job, so any previous .sig from the same type of build could be there.

Changed node context to be consistent with the other pipeline stage contexts, and cleared workspace prior to copying the GPG job artifacts.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -762,7 +762,9 @@ class Build {
                propagate: true,
                parameters: params
 
-           context.node('gpgsign') {
+           context.node('worker') {
+               // Remove any previous workspace artifacts
+               context.sh "rm -rf workspace/target/* || true"
                context.copyArtifacts(
                     projectName: "build-scripts/release/sign_temurin_gpg",
                     selector: context.specific("${signSHAsJob.getNumber()}"),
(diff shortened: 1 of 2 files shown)


andrew-m-leonard:
run tests

karianna (approved):
LGTM

sxa (changes requested):
This was done for security reasons. I do not approve this

zdtsw (approved):
LGTM

andrew-m-leonard:
> This was done for security reasons. I do not approve this

@sxa I can't see what the problem is, this is just the running of copyArtifact of the .sig files, which end up being published anyway... please explain?

sxa:
> > This was done for security reasons. I do not approve this
> 
> @sxa I can't see what the problem is, this is just the running of copyArtifact of the .sig files, which end up being published anyway... please explain?

Yeah ok I thought more than that had been changed here (I DM'd you on slack to try and discuss as I don't want to discuss any potential security issues in public). I'll remove my block

andrew-m-leonard:
run tests

andrew-m-leonard:
pr tester failed due to missing context.node("worker") prior to workspace clean up... fixed.

PR #370, opened 2022-07-19
fix: convert buildNumber to string for jenkins copyArtifacts do not accept int

zdtsw:
To fix: https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk8-pipeline/82/console

Ref: https://github.com/jenkinsci/workflow-support-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java#L135


karianna (approved):
LGTM

andrew-m-leonard (approved):

PR #340, opened 2022-06-27
fix: add recursive for rm so it won't fail because of folder

zdtsw:
fix error when pipeline cannot rm a folder
` rm: cannot remove 'workspace/target/AQAvitTaps': Is a directory`
e.g https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk19/job/jdk19-windows-x64-temurin/11/consoleFull

PR #371, opened 2022-07-22
fix: missing windows msi build for jre

zdtsw:
- we should not have a cleanup workspace before we do listArchives in buildWindowsInstaller this cause the windows installer job dont find a "jre" to match
	- we have moved to use worker node, not master or build-in so the cleanup is not needed as well 

Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/374

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -569,8 +569,8 @@ class Build {
     Run the Windows installer downstream jobs.
     We run two jobs if we have a JRE (see https://github.com/adoptium/temurin-build/issues/1751).
     */
-    private void buildWindowsInstaller(VersionInfo versionData) {
-        def filter = "**/OpenJDK*jdk_*_windows*.zip"
+    private void buildWindowsInstaller(VersionInfo versionData, String filter, String category) {
+        def nodeFilter = "${buildConfig.TARGET_OS}&&wix"
 
         def buildNumber = versionData.build
 
@@ -600,10 +600,10 @@ class Build {
                         context.string(name: 'PRODUCT_PATCH_VERSION', value: "${patch_version}"),
                         context.string(name: 'PRODUCT_BUILD_NUMBER', value: "${buildNumber}"),
                         context.string(name: 'MSI_PRODUCT_VERSION', value: "${versionData.msi_product_version}"),
-                        context.string(name: 'PRODUCT_CATEGORY', value: "jdk"),
+                        context.string(name: 'PRODUCT_CATEGORY', value: "${category}"),
                         context.string(name: 'JVM', value: "${buildConfig.VARIANT}"),
                         context.string(name: 'ARCH', value: "${INSTALLER_ARCH}"),
-                        ['$class': 'LabelParameterValue', name: 'NODE_LABEL', label: "${buildConfig.TARGET_OS}&&wix"]
+                        ['$class': 'LabelParameterValue', name: 'NODE_LABEL', label: "${nodeFilter}"]
                 ]
         context.copyArtifacts(
                 projectName: "build-scripts/release/create_installer_windows",
@@ -612,42 +612,6 @@ class Build {
                 fingerprintArtifacts: true,
                 target: "workspace/target/",
                 flatten: true)
-
-        // Check if JRE exists, if so, build another installer for it
-        listArchives().each({ file ->
-
-            if (file.contains("-jre")) {
-
-                context.println("We have a JRE. Running another installer for it...")
-                def jreinstallerJob = context.build job: "build-scripts/release/create_installer_windows",
-                        propagate: true,
-                        parameters: [
-                            context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
-                            context.string(name: 'UPSTREAM_JOB_NAME', value: "${env.JOB_NAME}"),
-                            context.string(name: 'FILTER', value: "**/OpenJDK*jre_*_windows*.zip"),
-                            context.string(name: 'PRODUCT_MAJOR_VERSION', value: "${versionData.major}"),
-                            context.string(name: 'PRODUCT_MINOR_VERSION', value: "${versionData.minor}"),
-                            context.string(name: 'PRODUCT_MAINTENANCE_VERSION', value: "${versionData.security}"),
-                            context.string(name: 'PRODUCT_PATCH_VERSION', value: "${patch_version}"),
-                            context.string(name: 'PRODUCT_BUILD_NUMBER', value: "${buildNumber}"),
-                            context.string(name: 'MSI_PRODUCT_VERSION', value: "${versionData.msi_product_version}"),
-                            context.string(name: 'PRODUCT_CATEGORY', value: "jre"),
-                            context.string(name: 'JVM', value: "${buildConfig.VARIANT}"),
-                            context.string(name: 'ARCH', value: "${INSTALLER_ARCH}"),
-                            ['$class': 'LabelParameterValue', name: 'NODE_LABEL', label: "${buildConfig.TARGET_OS}&&wix"]
-                        ]
-
-                context.copyArtifacts(
-                    projectName: "build-scripts/release/create_installer_windows",
-                    selector: context.specific("${jreinstallerJob.getNumber()}"),
-                    filter: 'wix/ReleaseDir/*',
-                    fingerprintArtifacts: true,
-                    target: "workspace/target/",
-                    flatten: true
-                )
-            }
-
-        })
     }
 
     /*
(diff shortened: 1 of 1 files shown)


sophia-guo:
For details https://adoptium.slack.com/archives/CLCFNV2JG/p1658500701043129

Code was changed https://github.com/adoptium/ci-jenkins-pipelines/pull/360/files

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -663,11 +663,9 @@ class Build {
 
         context.node('worker') {
             context.stage("installer") {
-                // Ensure master context workspace is clean of any previous archives
-                context.sh "rm -rf workspace/target/* || true"
 
                 switch (buildConfig.TARGET_OS) {
-                    case "mac":     buildMacInstaller(versionData); break
+                    case "mac": context.sh 'rm -f workspace/target/*.pkg workspace/target/*.pkg.json workspace/target/*.pkg.sha256.txt'; buildMacInstaller(versionData); break
```
sophia-guo, line 668: In this case for windows previous archives will not be cleaned, which will be not clean the any previous workspace artifacts as PR https://github.com/adoptium/ci-jenkins-pipelines/pull/360/files expected. `Ensure node context workspace is clean prior to gpg signing `

sophia-guo: it's actually updated by https://github.com/adoptium/ci-jenkins-pipelines/pull/340/files

sophia-guo:
Instead of using listArchives() to check `worker` space https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L790-L799, jdk11u/job/jdk11u-windows-x64-temurin/89/execution/node/294/ws/workspace/target/ might just update the logic of https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L617-L619 to check if corresponding     `UPSTREAM_JOB_NUMBER` `UPSTREAM_JOB_NAME` has the jre file `**/OpenJDK*jre_*_windows*.zip`?

pipelines/build/common/openjdk_build_pipeline.groovy
zdtsw: yeah, the reason first  to add `rm -rf workspace/target/*` is because the code was running on the "master|built-in" (multiple nodes, than before was only one master)
then later it changed from "master|built-in" to "node" which should not have the problem.

zdtsw:
> Instead of using listArchives() to check `worker` space https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L790-L799, jdk11u/job/jdk11u-windows-x64-temurin/89/execution/node/294/ws/workspace/target/ might just update the logic of https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L617-L619 to check if corresponding `UPSTREAM_JOB_NUMBER` `UPSTREAM_JOB_NAME` has the jre file `**/OpenJDK*jre_*_windows*.zip`?

not really what how it is implemented inside the job "create_installer_windows"
but if "create_installer_mac" can support both jdk and jre without using filters to be called twice, 
we should have the logic builtup the same way for "create_installer_windows" too.

zdtsw:
old thread regarding why specifically calling create_installer_windows twice : https://github.com/adoptium/temurin-build/issues/1751

sophia-guo (dismissed):
For the current PR how the the workspace/target/* is cleaned on windows?

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -663,13 +627,19 @@ class Build {
...
             context.stage("installer") {
-                // Ensure master context workspace is clean of any previous archives
-                context.sh "rm -rf workspace/target/* || true"
 
                 switch (buildConfig.TARGET_OS) {
-                    case "mac":     buildMacInstaller(versionData); break
-                    case "windows": buildWindowsInstaller(versionData); break
-                    default: break
+                    case "mac":
+                        context.sh 'rm -f workspace/target/*.pkg workspace/target/*.pkg.json workspace/target/*.pkg.sha256.txt';
```
sophia-guo, line 633: If https://github.com/adoptium/ci-jenkins-pipelines/pull/371#issuecomment-1202674831 do the cleanup

Why mac need extra cleanup?

zdtsw: the function of `cleanWorkspaceBuildOutputAfter` was added by Morgan Davies on 2021.02.04 from 485a9775ec8a353da63abdd4e9a5cc2261139e93
the `"context.sh 'rm -f....."` for `case "mac"` was added by Adam Farley on 2021.01.07 9537297119aa13bc1a2a9e86ebf550b0428c93f6
it was the same case when linux as in the same switch-case without cleanup, till e18d9e9c59e6079b48b57701fcb40074775b727c remove it

gdams (approved):

karianna:
@sophia-guo just yours re review to go I think.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -663,13 +627,19 @@ class Build {
 
         context.node('worker') {
             context.stage("installer") {
-                // Ensure master context workspace is clean of any previous archives
-                context.sh "rm -rf workspace/target/* || true"
```
andrew-m-leonard, line 667: These workspace clear need to remain. the reason they have to be target/* is because a previous installer invocation on the "worker" node will potentially still have old artifacts there.
For example, even though say we are doing a Mac Installer, there maybe a previous Win Installer job .zip file there...

andrew-m-leonard (dismissed):

pipelines/build/common/openjdk_build_pipeline.groovy
zdtsw: should not https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L1463 be triggered for the cleanup as post step by previous run?

andrew-m-leonard: we should be using context.copyArtifacts() ....

zdtsw:
run tests quick 11

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -663,13 +627,27 @@ class Build {
...
+                    case "mac":
+                        context.sh "rm -rf workspace/target/* || true"
+                        buildMacInstaller(versionData);
+                        break
+                    case "windows":
+                        context.sh "rm -rf workspace/target/* || true"
+                        buildWindowsInstaller(versionData,"**/OpenJDK*jdk_*_windows*.zip", "jdk");
+                        // Check if JRE exists, if so, build another installer for it
+                        context.copyArtifacts(
+                            projectName: context.specific("${env.JOB_NAME}"),
```
andrew-m-leonard, line 641: remove the context.specific()'s ...

andrew-m-leonard (approved):
looks good now

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -663,13 +627,27 @@ class Build {
...
-                    case "windows": buildWindowsInstaller(versionData); break
-                    default: break
+                    case "mac":
+                        context.sh "rm -rf workspace/target/* || true"
+                        buildMacInstaller(versionData);
+                        break
+                    case "windows":
+                        context.sh "rm -rf workspace/target/* || true"
+                        buildWindowsInstaller(versionData,"**/OpenJDK*jdk_*_windows*.zip", "jdk");
+                        // Check if JRE exists, if so, build another installer for it
```
karianna, line 639: comment should be dropped lower?

karianna (dismissed):

pipelines/build/common/openjdk_build_pipeline.groovy
zdtsw: updated.

zdtsw:
job from pr-test: https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/build-test/job/jobs/job/jdk11u/job/jdk11u-windows-x64-temurin/67/

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -663,13 +627,28 @@ class Build {
 
         context.node('worker') {
             context.stage("installer") {
-                // Ensure master context workspace is clean of any previous archives
-                context.sh "rm -rf workspace/target/* || true"
```
sophia-guo, line 667: These workspace clear need to remain. What's the point of move this clear to case " mac" and case " windows" separately ( Line 633 and Line 637)? @zdtsw

zdtsw: no specific reason to have same line in both case.
unless we want to have regex to only delete certain file patterns in the future.

PR #382, opened 2022-08-23
fix: set timeout to 1hr for  smoke test

zdtsw:
This PR is to set 1hr as timeout  for smoketest stage
Fix: https://github.com/adoptium/ci-jenkins-pipelines/issues/383

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -299,24 +299,19 @@ class Build {
                             context.jobDsl targets: templatePath, ignoreExisting: false, additionalParameters: jobParams
                         }
                     }
-                    context.catchError {
-                        def smoketestJob = context.build job: jobName,
-                                propagate: false,
-                                parameters: [
-                                        context.string(name: 'SDK_RESOURCE', value: "upstream"),
-                                        context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
-                                        context.string(name: 'UPSTREAM_JOB_NAME', value: "${env.JOB_NAME}"),
-                                        context.string(name: 'JDK_VERSION', value: "${jobParams.JDK_VERSIONS}"),
-                                        context.string(name: 'LABEL_ADDITION', value: additionalTestLabel),
-                                        context.booleanParam(name: 'KEEP_REPORTDIR', value: buildConfig.KEEP_TEST_REPORTDIR),
-                                        context.string(name: 'ACTIVE_NODE_TIMEOUT', value: "${buildConfig.ACTIVE_NODE_TIMEOUT}"),
-                                        context.booleanParam(name: 'DYNAMIC_COMPILE', value: true)]
-                        def smokeResult = smoketestJob.getResult()
-                        if (smokeResult != 'SUCCESS') { // failed or unstable or aborted
-                            context.error("${jobName} result is ${smokeResult}")
-                            currentBuild.result = "${smokeResult}"
-                        }
-                    }
+                }
+                context.catchError {
+                    context.build job: jobName,
+                            propagate: false,
+                            parameters: [
+                                    context.string(name: 'SDK_RESOURCE', value: "upstream"),
+                                    context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
+                                    context.string(name: 'UPSTREAM_JOB_NAME', value: "${env.JOB_NAME}"),
+                                    context.string(name: 'JDK_VERSION', value: "${jobParams.JDK_VERSIONS}"),
+                                    context.string(name: 'LABEL_ADDITION', value: additionalTestLabel),
+                                    context.booleanParam(name: 'KEEP_REPORTDIR', value: buildConfig.KEEP_TEST_REPORTDIR),
+                                    context.string(name: 'ACTIVE_NODE_TIMEOUT', value: "${buildConfig.ACTIVE_NODE_TIMEOUT}"),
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
The comment indicates the intention was to only run AQA if SomeTest succeeds?
https://github.com/adoptium/ci-jenkins-pipelines/blob/6be5a31e9f712c4638efa29df5c5c1ce445e26cb/pipelines/build/common/openjdk_build_pipeline.groovy#L281

andrew-m-leonard (commented):
Can we confirm this behavior is correct with @smlambert ?

zdtsw:
yeah, i think if we want to keep "only good smoke then start run AQA" I can remove the "wait: false", also set the pipeline result to smoketest result.
at the moment, it is not doing the correct thing, so the check for smoketest result is still needed.

smlambert:
> Run smoke tests, which should block the running of downstream test jobs if there are failures.

That is ultimately the behaviour we want.  We had not yet enabled this behaviour, because there were still many smoke tests failing at the time of the original change and were asked not to block other testing due to them failing.

Reminding that the smoke tests from the temurin-build repo are checking "did we build the right thing", if they fail, it means we did not build the right thing and we should not waste resources on extra testing for it.

zdtsw:
update the logic of this PR to: gate aqa-tests by the result of smoke if it is not successful.

llxia:
I am not sure if all the smoke tests are green. The following test failed for Openj9 at Adopt, but passed for Semeru.
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-aarch64-openj9_SmokeTests/190/

llxia:
FYI @AdamBrousseau

karianna (approved):
Looks sane as an isolated change.  lots of Whitespace changes again though

zdtsw:
> Looks sane as an isolated change. lots of Whitespace changes again though

to enable "hide whitespace"? 
Screenshot from 2022-08-24 08-05-40

zdtsw:
run tests quick 11

zdtsw:
reverted the part to check result of smoketest, only set timeout to 1hr.
so we can proceed this PR (if we are not ready to use smoketest result to gate pipeline)

llxia:
May I know why we do not use the existing parameter TIME_LIMIT and introduce a new parameter SMOKETEST_TIMEOUT?

smlambert (changes requested):
Please take a look at the suggestion in the issue this fixes for a preferred way to change this timeout (https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L143).

zdtsw:
Basically I see two ways to add timeout

1. in pipeline set timeout over stage then call downstream smoketest job  ( as in this PR) which is similar to coding style to other stage https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L74 this introduces a new constant


2. in the smoketest job itself by re-using TIME_LIMIT, which is similar to https://github.com/adoptium/aqa-tests/pull/3928/files?diff=split&w=1 then in the pipeline add `jobParams.put('TIME_LIMIT', "1")` 

Or, did I miss something as preferred way?

sophia-guo (approved):

smlambert:
Somehow I imagined to use the known pre-existing variable TIME_LIMIT and add an extra put (`jobParams.put("TIME_LIMIT", 1)` in getSmokeTestJobParams()

```
     def getSmokeTestJobParams() {
        def jobParams = getCommonTestJobParams()
        jobParams.put('LEVELS', "extended")
        jobParams.put('GROUPS', "functional")
        jobParams.put('TEST_JOB_NAME', "${env.JOB_NAME}_SmokeTests")
        jobParams.put('BUILD_LIST','functional/buildAndPackage')
        def useAdoptShellScripts = Boolean.valueOf(buildConfig.USE_ADOPT_SHELL_SCRIPTS)
        def vendorTestRepos = ((String)ADOPT_DEFAULTS_JSON['repository']['build_url']).minus(".git")
        def vendorTestBranches = ADOPT_DEFAULTS_JSON['repository']['build_branch']
        def vendorTestDirs = ADOPT_DEFAULTS_JSON['repository']['test_dirs']
        if (!useAdoptShellScripts) {
            vendorTestRepos = ((String)DEFAULTS_JSON['repository']['build_url']).minus(".git")
            vendorTestBranches = DEFAULTS_JSON['repository']['build_branch']
            vendorTestDirs = DEFAULTS_JSON['repository']['test_dirs']
        }                     
        jobParams.put("VENDOR_TEST_REPOS", vendorTestRepos)
        jobParams.put("VENDOR_TEST_BRANCHES", vendorTestBranches)
        jobParams.put("VENDOR_TEST_DIRS", vendorTestDirs)
        jobParams.put("TIME_LIMIT", 1)
        return jobParams
    }
```

smlambert (commented):
Removing my request changes to leave to others for review.  While I prefer to not introduce yet another constant, it does appear to be a method used in the top-level pipeline.

zdtsw:
> I also prefer not to introduce another new parameter. Especially, there is an existing parameter that serves the purpose. Adding an additional parameter to resolve a similar issue will cause extra maintenance and confusion.

Do you feel it is better to have https://github.com/adoptium/aqa-tests/pull/3928/files?diff=split&w=1 and https://github.com/adoptium/ci-jenkins-pipelines/pull/394 together as solution?

zdtsw:
if we all agree on the solution in https://github.com/adoptium/ci-jenkins-pipelines/pull/394 then I can have that one open and close this one? @llxia @smlambert @sophia-guo

zdtsw:
have this one closed and continue with https://github.com/adoptium/ci-jenkins-pipelines/pull/394

PR #386, opened 2022-08-25
re-enable windows aarch64 pipelines

gdams:
A stable Azure machine has been added for testing now so this should be good to re-enable (https://github.com/adoptium/infrastructure/issues/2708)

Once I've confirmed that everything is working well I'll provision a second machine for redundancy

### DIFF
--- a/pipelines/jobs/configurations/jdk11u.groovy
+++ b/pipelines/jobs/configurations/jdk11u.groovy
@@ -3,6 +3,7 @@ targetConfigurations = [
         "x64Linux"      : [    "temurin",    "openj9",    "dragonwell",    "corretto",    "bisheng",    "fast_startup"],
         "x64AlpineLinux": [    "temurin"                            ],
         "x64Windows"    : [    "temurin",    "openj9",    "dragonwell"            ],
+        "aarch64Windows": [    "temurin"                            ],   
         "x32Windows"    : [    "temurin"                            ],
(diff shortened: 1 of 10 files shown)


zdtsw (approved):

karianna:
In theory with the latest VS for Windows on ARM we could try native compiling?

gdams:
My thought process was the re-enable what we knew to be working and then we can loop back round and try to enable a native build for JDK17+ once we know the azure machines are stable.

karianna (approved):
LGTM

smlambert:
Test support should be added ahead of the enablement of testing from the build pipelines when new platforms are added.  Please disable testing on this platform until test support is added for it.

gdams:
@smlambert I believe that Microsoft added Windows aarch64 support to the aqa-tests repo? If there is something missing I'll raise a PR.

smlambert:
False, that platform has not been [added](https://github.com/adoptium/TKG/blob/master/resources/buildPlatformMap.properties) to date.  

Though I did create those [instructions](https://github.com/adoptium/aqa-tests/wiki/Adding-AQAvit-support-for-a-New-Platform) and shared them with Joe B. in the testing-aqavit Slack channel so that he'd know how to start.  

There may be complications due to no cygwin present and appears like wsl has 'differences' that will need to be handled.  Last time this was attempted, we had no machine that we could log on to and try, so the [issue](https://github.com/adoptium/aqa-tests/issues/2224) was parked.

Issue #389, opened 2022-08-29
weekly pipeline failed even downstream jobs are all green


zdtsw:
from https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk11-pipeline/85/
with error:
`Also:   org.jenkinsci.plugins.workflow.steps.MissingContextVariableException: Required context class hudson.FilePath is missing
Perhaps you forgot to surround the code with a step that provides this, such as: node`

PR #390, opened 2022-08-29
fix: add missing node for job to do copyartifact for scriptive Jenkins pipeline

zdtsw:
weekly pipeline seems is the only jenkins job we are using scriptive syntax
before we added copyartifact step, it was only used to trigger downstream jobs
after we add copyartifact, it need to be allocated onto a node to run the step.

Fix: https://github.com/adoptium/ci-jenkins-pipelines/issues/389

### DIFF
--- a/pipelines/build/common/weekly_release_pipeline.groovy
+++ b/pipelines/build/common/weekly_release_pipeline.groovy
@@ -13,57 +13,58 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+node{
+    stage("Submit Release Pipelines") {
+        // Map of <variant> : <scmRef>
+        def Map<String, String> scmRefs = new JsonSlurper().parseText("${params.scmReferences}") as Map
 
-stage("Submit Release Pipelines") {
-    // Map of <variant> : <scmRef>
-    def Map<String, String> scmRefs = new JsonSlurper().parseText("${params.scmReferences}") as Map
+        // Map of <platform> : [<variant>,<variant>,..]
+        def Map<String, List<String>> targetConfigurations = new JsonSlurper().parseText("${params.targetConfigurations}") as Map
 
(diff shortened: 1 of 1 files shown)


sophia-guo:
Might just use controller to copyartifacts to release limited node resources. Similar to https://github.com/adoptium/aqa-tests/pull/3941

zdtsw:
> Might just use controller to copyartifacts to release limited node resources. Similar to [adoptium/aqa-tests#3941](https://github.com/adoptium/aqa-tests/pull/3941)

Yes! ths missing piece here is just the "node" part  line16/70 (easy to review the change if "hide whitespace")

karianna (approved):
LGTM

andrew-m-leonard (approved):

andrew-m-leonard:
run tests

zdtsw:
run tests

sophia-guo:
> Might just use controller to copyartifacts to release limited node resources. Similar to [adoptium/aqa-tests#3941](https://github.com/adoptium/aqa-tests/pull/3941)

My point is only running step copyArtifacts on node ( i.e. 'worker'), which means all other steps shouldn't run on node.  Especially triggering the parallel job and waiting for the job results should not run on node('worker'). Otherwise the executor of the node 'worker' will be hold for the whole job running, which I believe might take long enough for some variant and platform. @andrew-m-leonard @zdtsw

zdtsw:
> > Might just use controller to copyartifacts to release limited node resources. Similar to [adoptium/aqa-tests#3941](https://github.com/adoptium/aqa-tests/pull/3941)
> 
> My point is only running step copyArtifacts on node ( i.e. 'worker'), which means all other steps shouldn't run on node. Especially triggering the parallel job and waiting for the job results should not run on node('worker'). Otherwise the executor of the node 'worker' will be hold for the whole job running, which I believe might take long enough for some variants and platforms. @andrew-m-leonard @zdtsw

great idea! i did not get it but now i think i know what you meant
i will do another commit to move the node to only for copyArtifacts step

PR #391, opened 2022-09-01
Set SDK_RESOURCE=upstream for aqa-tests

### LINKED ISSUE #392: Grinder should use the same parameter value from upstream job
**Describe the bug**
Trying to run grinder from test job, but grinder trying to download "nightly" jdk from API VS original upstream job is copying jdk from build artifact
This could give the false test result if jdk is not the one from last build. 
Since we have some problems to download the latest ones from jdk19/20 therefor grinder failed, 
But for other jdk versions, grinder wont failed to download jdk so the test result might not be accurate.  

**To Reproduce**
original upstream job https://ci.adoptopenjdk.net/job/Test_openjdk20_hs_sanity.functional_x86-64_alpine-linux/11/consoleFull
click link from console to run grinder https://ci.adoptopenjdk.net/job/Grinder/5564/console

**Expected behavior**
grinder pre-fill value 
Screenshot from 2022-09-01 07-53-29
because in original upstream job it does not set SDK_RESOURCE, then grinder fall back to the first item from dropdownlist "nightly"

**Screenshots**
Screenshot from 2022-09-01 07-54-46

both jobs set correct upstream job information
Screenshot from 2022-09-01 07-56-04

sophia-guo:
Close #392 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -416,6 +416,7 @@ class Build {
                                             parameters: [
                                                 context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
                                                 context.string(name: 'UPSTREAM_JOB_NAME', value: "${env.JOB_NAME}"),
+                                                context.string(name: 'SDK_RESOURCE', value: "upstream"),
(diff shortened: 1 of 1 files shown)


smlambert (approved):

karianna (approved):
LGTM

PR #393, opened 2022-09-02
improve: move "node" block to only on the copyartifacts step to reduce resource lock by long running job

zdtsw:
From comments: https://github.com/adoptium/ci-jenkins-pipelines/pull/390#issuecomment-1234599140
this change can reduce resource/"worker" node be locked with executor's number during a long weekly pipeline run.

(use Hide Whitespace for better review experience)

Also fix wrong filter
from openjdkXX pipeline copy up to weekly job, does not have workspace/target (this is for downstream build job)

Fix: https://github.com/adoptium/ci-jenkins-pipelines/issues/389


sophia-guo (approved):
LGTM

karianna (approved):
LGTM

Issue #383, opened 2022-08-23
Reduce TIME_LIMIT for smoke tests


smlambert:
The current set of smoke tests should only take a few minutes to run, and taking into account that a few more smoke tests for temurin-build should get added, it is reasonable to reduce the TIME_LIMIT for those tests from the default 10hrs, to something much smaller, like 1 hr or 30 min.

This can be done by explicitly setting TIME_LIMIT onto the set of smoketest parameters here:
https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L143

sophia-guo:
As mentioned in https://github.com/adoptium/ci-jenkins-pipelines/pull/403#issue-1367401699 all related fix will make sure the new smoketests job be created with reduced TIME_LIMIT=1. However for existing ones there is no way for fix it as the jobs were created with TIME_LIMIT=10 as a  string not a variable. 

That is to fully fix this issue we need to regenerate all existing smoke tests jobs. @smlambert

smlambert:
Right, so we can run a nightly or weekly pipeline with the AQA_AUTO_GEN flag to true.

zdtsw:
this should work now after #403 got merged
![Screenshot from 2022-09-13 08-15-37](https://user-images.githubusercontent.com/915053/189824401-e7b4b132-6b07-4e5e-9bbd-b93b88b08fb4.png)  from https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk19/job/jdk19-alpine-linux-x64-temurin_SmokeTests/

sophia-guo:
yeah, didn't realized that [TIME_LIMIT is set by variable ](https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/testJobTemplate#L371). 

Agreed that we can keep the existing job configurations as is (default 10). For non-existing ones I don't see the reason that we can't generate the job with default 1 unless smoketests will be expected to run longer than 1 hour. I'm fine with either decision.

PR #394, opened 2022-09-02
feat: add default timeout to 1hr for smoketest

zdtsw:
Fix: https://github.com/adoptium/ci-jenkins-pipelines/issues/383 

To have this solution need to have changes in aqa-tests as well


sophia-guo (approved):

llxia (approved):
LGTM

sophia-guo (approved):

Issue #301, opened 2022-05-18
Ability to run a Reproducible Build comparison Jenkins job to Verify two builds are identical


andrew-m-leonard:
To aid verification of reproducible builds, it would be nice to have a Jenkins job that takes parameters like:
- JDK 1 download "URL"
- JDK 2 download "URL"
- Platform (eg.MacOS_aarch64)

The job then downloads both JDKs and performs a strict reproducible build comparison on them to indicate if they are identical.
"Platform" is required as reproducible build verification will vary based on platform (eg.Mac binaries need signing Signature removing before compare).

zdtsw:
i do not quite follow this issue. 
a jenkins job to download same JDK versions but from different URL(?) then use them build same code respectively and compare the outcome?

zdtsw:
in the case of URL1 and URL2, do you have one example? 
I can only think of temurin*-binaries git repo.  When the upstream code has no change so the nightly build could produce the same binary.

andrew-m-leonard:
So reproducible builds will be "RELEASE=true" type builds, as that enables the --with-source-date=version and the reproducible flag.
So doing a release build twice for example, the URL1 could be to the binaries repo JDK.tar.gz and URL2 could be a re-build of the same release in a Jenkins job artifact.
Equally it could be for test/debug purpose so URL could be to where ever you place your JDK.tar.gz, eg.google docs

smlambert:
So this job will take 2 parameters, URL1, URL2 both of which would be expected to point to a build that has been run with RELEASE=true (but this will be an expectation that the user knows this and/or reads the fine print under the parameter.

As an additional note, it is similar to how test jobs can take CUSTOMIZED_SDK_URL, this includes if you have build something locally on your laptop and then use UploadFile job to host it, to pass the URL1 or URL2 to this job.

zdtsw:
Sum up after talk  with @andrew-m-leonard :
- release(reproducible) build is only applied on jdk19 (and will be back ported to jdk17 later), focus now is on jdk19(head) builds
- in reality we want to build twice on the same SHA from upstream(openjdk source) which generate two builds' artifacts correspondingly for downloadable as URL1 and URL2
- Preferably this should be a new dedicated jenkins job, which we can run manually when needed but mostly be triggered as part of the  weekly-openjdkX-pipeline in the future
[...]

PR #401, opened 2022-09-06
Correct copy directory

### LINKED ISSUE #400: weekly pipeline failed with 'ERROR: Failed to copy artifacts from openjdk11-pipeline with filter: workspace/target/*'
All weekly pipeline failed with following message even all jdk pipeline builds are green.

```
ERROR: Failed to copy artifacts from openjdk11-pipeline with filter: workspace/target/*
Finished: FAILURE
```
https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk11-pipeline/86/

sophia-guo:
Fix #400 


sophia-guo:
Would also be better if the artifacts can be narrowed down. Feels like in original issue https://github.com/adoptium/ci-jenkins-pipelines/issues/301 only jdk jars are needed? This can be done separately once confirmed.

smlambert (approved):

PR #402, opened 2022-09-07
fix: only copy up tar.gz and zip file as needed for comparison

zdtsw:
- to only copy tar.gz and zip file, to exclude others msi pkg json etc
	 - have jdk jre testimage static all copied up
	 - flat target to put everything in the same folder since each name is unique 

This one will be test out first in weekly openjdk19/20 to see if it does work
waiting https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk20-pipeline/ to be run on 11th morning


sophia-guo:
Part of https://github.com/adoptium/ci-jenkins-pipelines/issues/301

zdtsw:
test on https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk17-pipeline/84/console does filter out the other non-tar non-zip file, but still the last part need node{} block to do archive

karianna (approved):
LGTM

sophia-guo (approved):

Issue #406, opened 2022-09-13
Nightly build stats build is not working on jdk8


zdtsw:
**What are you trying to do?**
we do have build 2 days ago, but the result saying jdk8 last published 7 days ago
this is because we made an EA build for jdk8 and API sort it by name which became the first element to return https://github.com/adoptium/ci-jenkins-pipelines/blob/master/tools/nightly_build_and_test_stats.groovy#L50

so jdk8u352-b05-ea will be used all the time for jdk8 nightly build status which is incorrect.

**Expected behaviour:**
Get correct status from job https://ci.adoptopenjdk.net/view/Tooling/job/nightlyBuildAndTestStats_temurin/804/console
```
[
    {
        "binaries": [
            {
                "architecture": "x64",
                "download_count": 395,
                "heap_size": "normal",
                "image_type": "jdk",
                "jvm_impl": "hotspot",
                "os": "linux",
                "package": {
                    "checksum": "2bf66c3e1c010b7df3ecfde53f4d280fb6352d0617393e4711756f3326d9a1de",
                    "checksum_link": "https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u352-b05-ea/OpenJDK8U-jdk_x64_linux_hotspot_8u352b05_ea.tar.gz.sha256.txt",
                    "download_count": 395,
                    "link": "https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u352-b05-ea/OpenJDK8U-jdk_x64_linux_hotspot_8u352b05_ea.tar.gz",
                    "metadata_link": "https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u352-b05-ea/OpenJDK8U-jdk_x64_linux_hotspot_8u352b05_ea.tar.gz.json",
                    "name": "OpenJDK8U-jdk_x64_linux_hotspot_8u352b05_ea.tar.gz",
                    "signature_link": "https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u352-b05-ea/OpenJDK8U-jdk_x64_linux_hotspot_8u352b05_ea.tar.gz.sig",
                    "size": 102936760
                },
                "project": "jdk",
                "scm_ref": "jdk8u352-b05_adopt",
                "updated_at": "2022-09-05T18:43:31Z"
            }
        ],
        "download_count": 6374,
        "id": "RE_kwDOFjpgDM4EjHkm.7R859jy0sVmsrw==",
[...]

PR #407, opened 2022-09-13
fix: explicit set nightly status to ignore first return element on jdk8

zdtsw:
fix https://github.com/adoptium/ci-jenkins-pipelines/issues/406


jerboaa (changes requested):
I don't think that's the right fix. See: https://github.com/adoptium/ci-jenkins-pipelines/issues/406#issuecomment-1245355957

jerboaa (approved):
LGTM.

karianna (approved):
LGTM

PR #415, opened 2022-09-19
Pull riscv64 sources from default repository

luhenry:
Since https://openjdk.org/jeps/422 has been merged in JDK 19, we don't need to pull from the riscv-port repo anymore and we can pull from the default repo instead.

### DIFF
--- a/pipelines/jobs/configurations/jdk19_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk19_pipeline_config.groovy
@@ -171,7 +171,9 @@ class Config19 {
                 os                   : 'linux',
                 arch                 : 'riscv64',
                 configureArgs        : '--enable-dtrace --with-native-debug-symbols=none',
-                buildArgs            : '-r https://github.com/openjdk/riscv-port -b riscv-port --custom-cacerts false --disable-adopt-branch-safety --create-sbom',
+                buildArgs           : [
+                        "temurin"   : '--create-jre-image --create-sbom'
+                ],
                 test                : [
                         nightly: ['sanity.openjdk'],
                         weekly : ['sanity.openjdk', 'sanity.system', 'extended.system', 'sanity.perf']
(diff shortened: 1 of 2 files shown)


karianna (approved):
LGTM - holding off merging until after release week though.

karianna:
@luhenry Are you also able to sign the Eclipse CLA?

PS:  Nice to see you here!

luhenry:
@karianna I'm working on signing the Eclipse CLA. I'll come back to this PR as soon as it's done.

gdams:
@luhenry here's the docs if you need help https://adoptium.net/docs/eca-sign-off

pipelines/jobs/configurations/jdk19_pipeline_config.groovy
```diff
@@ -171,7 +171,9 @@ class Config19 {
                 os                   : 'linux',
                 arch                 : 'riscv64',
                 configureArgs        : '--enable-dtrace --with-native-debug-symbols=none',
```
gdams, line 173: do you know if we still need to pass `--with-native-debug-symbols=none` to the configure command? That doesn't seem quite right

luhenry: We don’t in upstream OpenJDK GHA. Not sure why it’s here in the first place though. Also, getting the debug symbols is very handy.

gdams: ping @sxa who likey has more insight here

luhenry:
I was able to sign the ECA. It should be now good to review.

pipelines/jobs/configurations/jdk19_pipeline_config.groovy
luhenry: From discussion on Slack, it was disabled due to the extra cost of storing/transferring the data in and out of the bots. Let's reenable it and disable it if/when it breaks.

sxa (commented):
LGTM - thanks. I've let the workflows run now so if they pass I'll approve and merge

pipelines/jobs/configurations/jdk19_pipeline_config.groovy
sxa: Now changed - resolving

sxa (approved):

Issue #397, opened 2022-09-05
Ability to run a Reproducible Build comparison Jenkins job to Verify two builds are identical 


zdtsw:
This is the top epic which should include works for different OS:
1. Linux: https://github.com/adoptium/ci-jenkins-pipelines/issues/301
2. Windows
3. Mac

Description for this epic:
To create Jenkins job which does comparision work for the compressed file we create after build.
read https://github.com/adoptium/ci-jenkins-pipelines/issues/301 for more discussion

zdtsw:
@andrew-m-leonard 
job https://ci.adoptopenjdk.net/view/ReproducibleBuild/job/reproducible_comparision is result from #301 
from run result https://ci.adoptopenjdk.net/view/ReproducibleBuild/job/reproducible_comparision/2/console , expect it shows these two tar file of linux x64 are identical (same scm_ref from jdk17, same build_ref from temurin-build, can check detail sbom.json from GH release artifacts), but it says files are not the same.

Offline download these 2 tar files into folder `first` and `second`, and did a comparison on each's `jdk-17.0.5+7/bin/jar` 
used different tools to verify result
```
>cmp first/jdk-17.0.5+7/bin/jar second/jdk-17.0.5+7/bin/jar
first/jdk-17.0.5+7/bin/jar second/jdk-17.0.5+7/bin/jar differ: byte 8295, line 4
```
```
>xxd second/jdk-17.0.5+7/bin/jar > second.txt
>xxd first/jdk-17.0.5+7/bin/jar > first.txt
>diff second.txt first.txt 
519c519
< 00002060: 3032 3231 3030 3432 3330 3500 2d4a 2d6d  02210042305.-J-m
---
> 00002060: 3032 3231 3030 3530 3732 3900 2d4a 2d6d  02210050729.-J-m


```

zdtsw:
ah, I might know the reason. the build has `--with-source-data=updated`.

zdtsw:
https://ci.adoptopenjdk.net/view/ReproducibleBuild/job/reproducible_comparision/5/console is from two tar files both from weekly-build and same scm_ref
```
diff -q -r url1Dir1 url1Dir2
Files url1Dir1/jdk-17.0.5+7/lib/server/classes.jsa and url1Dir2/jdk-17.0.5+7/lib/server/classes.jsa differ
[...]

PR #416, opened 2022-09-27
fix: do not fail job if not matching file and correct archive path

zdtsw:
PR for two parts:
in https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk8-pipeline/89/console with error `ERROR: Failed to copy artifacts from openjdk8-pipeline with filter: **/*.tar.gz, **/*.zip` because corretto failed https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-x64-corretto/726/ so it does not generate artifacts. pipeline should fail in this case, but should not fail if something wrong when copy up artifact

in https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk18-pipeline/64/console with error `RROR: No artifacts found that match the file pattern "workspace". Configuration error?` because archiveArtifacts require files pattern not only dir.

Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/397

### DIFF
--- a/pipelines/build/common/weekly_release_pipeline.groovy
+++ b/pipelines/build/common/weekly_release_pipeline.groovy
@@ -55,7 +55,8 @@ stage("Submit Release Pipelines") {
                                 filter: '**/*.tar.gz, **/*.zip',
                                 fingerprintArtifacts: true,
                                 target: "workspace",
-                                flatten: true
+                                flatten: true,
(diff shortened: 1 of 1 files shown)


andrew-m-leonard (approved):

karianna (approved):
LGTM

PR #424, opened 2022-10-03
Disable remote trigger if tests are disabled

sophia-guo:
Close #422 

Signed-off-by: Sophia Guo <sophia.gwf@gmail.com>

PR #441, opened 2022-10-10
chore: prep for 2022 Oct CPU release to disable nightly test

zdtsw:
Disable Nightly test for 2022 Oct CPU release

### DIFF
--- a/pipelines/defaults.json
+++ b/pipelines/defaults.json
@@ -33,7 +33,7 @@
         "downstream"         : "pipelines/build/common/openjdk_build_pipeline.groovy"
     },
     "testDetails"            : {
-        "enableTests"        : true,
+        "enableTests"        : false,
         "nightlyDefault"     : [
             "sanity.openjdk",
             "sanity.system",


andrew-m-leonard (approved):

zdtsw:
now i start to think if this can be done now.....
in https://github.com/adoptium/ci-jenkins-pipelines/pull/424 if we set enableTests = false, then coming weekly wont triggere remote TCK jobs.

karianna (approved):
LGTM

zdtsw:
> now i start to think if this can be done.... in #424 if we set enableTests = false, then coming weekly wont triggere remote TCK jobs. and even the release build wont do the remote triggere.
> 
> So if we want to stop nightly run (enableTest=false) we can only merge this PR right after we have build our release binaries for all supported jdk versions.

after discussion, we should have this one merge now
when we run weekly or the official release we can set the input to False for the build

sophia-guo:
> when we run weekly or the official release we can set the input to False for the build

For the official release we can set the input to **true** for the build

PR #444, opened 2022-10-11
feat: add support to ignore classes_nocoops.jsa file on Linux to support reproducible builds

### LINKED ISSUE #435: Exclude  "classes_nocoops.jsa" file from tar-ball when doing reproducible comparison
**What are you trying to do?**
https://ci.adoptopenjdk.net/view/ReproducibleBuild/job/reproducible_comparision/ job when doing comparison with "diff" command should exclude "release" file which has build version info. etc are not identical
update: classes_nocoops.jsa is known non-deterministic issue at the moment. so it should be in the exclude list too. 

**Expected behaviour:**
comparison job should not fail because files in the excluded list are different from 2 tarball

**Observed behaviour:**
 "classes_nocoops.jsa" should not be listed
```
 diff -q -r url1Dir1 url1Dir2
Files url1Dir1/jdk-17.0.5+7/lib/server/classes.jsa and url1Dir2/jdk-17.0.5+7/lib/server/classes.jsa differ
Files url1Dir1/jdk-17.0.5+7/lib/server/classes_nocoops.jsa and url1Dir2/jdk-17.0.5+7/lib/server/classes_nocoops.jsa differ
Files url1Dir1/jdk-17.0.5+7/release and url1Dir2/jdk-17.0.5+7/release differ
```
**Any other comments:**
see https://github.com/adoptium/ci-jenkins-pipelines/issues/397#issuecomment-1269517946

sakshi1215:
Signed-off-by: sakshi1215 <sakshi.patle121@gmail.com>

fixes #435

### DIFF
--- a/tools/reproduce_comparison/Jenkinsfile
+++ b/tools/reproduce_comparison/Jenkinsfile
@@ -1,5 +1,5 @@
 /*
-Jenkins job does comparision for two different build artifacts
+Jenkins job does comparison for two different build artifacts
 If they are identical, build pass; or fail
 */
 
@@ -9,6 +9,7 @@ pipeline {
         choice(choices: ['linux', 'mac', 'windows'], name: 'platform', description:'select release on different platform')
         string(name: 'URL1', defaultValue: '', description: 'URL to one version of the build artifacts\ne.g.: https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk18-pipeline/lastSuccessfulBuild/artifact/target/linux/x64/temurin/OpenJDK18U-jdk_x64_linux_hotspot_2022-06-11-23-30.tar.gz')
         string(name: 'URL2', defaultValue: '', description: 'URL to the other version of the build artifacts\ne.g.: https://github.com/adoptium/temurin19-binaries/releases/download/jdk-2022-06-09-19-11-beta/OpenJDK-jdk_x64_linux_hotspot_2022-06-09-03-31.tar.gz')
+        string(name: 'excludeFile', defaultValue: 'classes_nocoops'. description: 'File name to be excluded while doing the comparison')
     }
 
     stages {
@@ -37,7 +38,7 @@ pipeline {
                         if (params.platform != 'linux') {
                             "prep${platform}" ('url1Dir1', 'url1Dir2')
                         }
-                        def retVal = sh returnStatus: true, script: 'diff -q -r url1Dir1 url1Dir2'
+                        def retVal = sh returnStatus: true, script: 'diff -q -r url1Dir1 url1Dir2 --exclude=${params.excludeFile}'
                         if (retVal != 0) {
                             currentBuild.result = 'FAILURE'
                             error 'Error: two builds are not the same!'
@@ -66,7 +67,7 @@ def prepmac(String jdk1, String jdk2, String localCert='defaultpath') {
 }
 
 /*
-TODO: placeholder for windows comparision
+TODO: placeholder for windows comparison
 def prepwindows(String jdk1, String jdk2, ...) {
    sh "./tools/reproduce_comparison/compareWindows.sh ..."
 }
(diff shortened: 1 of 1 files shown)


sakshi1215:
@zdtsw can you please review this?

tools/reproduce_comparison/Jenkinsfile
```diff
@@ -37,7 +38,7 @@ pipeline {
                         if (params.platform != 'linux') {
                             "prep${platform}" ('url1Dir1', 'url1Dir2')
                         }
-                        def retVal = sh returnStatus: true, script: 'diff -q -r url1Dir1 url1Dir2'
+                        def retVal = sh returnStatus: true, script: 'diff -q -r url1Dir1 url1Dir2 --exclude=${params.excludeFile}'
```
zdtsw, line 41: We do not need to introduce input as Jenkins parameters.
just make a hardcoded ArrayList is enough in this case. Becuse this is only running on linux, in other two OSs it is not clear how it should be done.
Until we want to create differnet jobs per different OS might be useful to take in user input in different cases.

One more thing, we need to make it work with multiple files.
At the moment, the only one is classes_nocoops.jsa but likely more will be added into the list.

And the file to be excluded is called `classes_nocoops.jsa` not `classes_nocoops`

sakshi1215: @zdtsw `classes_nocoops` is a pattern, so all files matching that pattern will be excluded. Diff doesn't support taking exclude input as a list of arrays so if we have to work with multiple files, we will need to append `--exclude` multiple times.

zdtsw: We only want to exclude `classes_nocoops.jsa` if there is another one match `classes_nocoops` we do not want to exclude it unless we are certain. Otherwise the requirement for comparison is inaccurate. 

As for the description of this issue is to have a list of it: originally two files `release` and `classes_nocoops.jsa` both should be excluded, then, `release` is removed so the title was updated with one `classes_nocoops.jsa`
Suggestion from comment is just an example for "good first issue" to start with, not as instruction. 
You need to add a loop and figure out how to handle say more than one file in this case.

sakshi1215: Understood, will add support to provide input by `,` delimiter & then iterate over the strings.

sakshi1215:
Screenshot 2022-10-12 at 7 45 13 PM
@zdtsw can you please check again, I have verified that my script works well with the `;` delimiter.

tools/reproduce_comparison/Jenkinsfile
sakshi1215: @zdtsw @karianna I have made the requested change!

tools/reproduce_comparison/Jenkinsfile
```diff
@@ -37,7 +38,12 @@ pipeline {
                         if (params.platform != 'linux') {
                             "prep${platform}" ('url1Dir1', 'url1Dir2')
                         }
-                        def retVal = sh returnStatus: true, script: 'diff -q -r url1Dir1 url1Dir2'
+                        excludeString=''
```
zdtsw, line 41: hum. I think here is groovy file not a shell script.

some snippet like this (not verified with syntax)
```
def excludeStrings = "file1;file2"
def excludeFlags =""
def excludeStringList = excludeStrings.split(';')
for(String excludeString : excludeStringList )
  excludeFlags="${excludeFlags} --exclude=${excludeString}"
​println "DEBUG: ${excludeFlags}​​​​​​​​​​​​​​​​"
def retVal = sh returnStatus: true, script: 'diff -q -r url1Dir1 url1Dir2 ${excludeFlags}'
```

sakshi1215: Ugh, I was honestly getting a lil confused since its been months I have not touched jenkins, let me fix it.

sakshi1215: Done @zdtsw please check again

zdtsw (approved):

tools/reproduce_comparison/Jenkinsfile
```diff
@@ -37,7 +38,12 @@ pipeline {
                         if (params.platform != 'linux') {
                             "prep${platform}" ('url1Dir1', 'url1Dir2')
                         }
-                        def retVal = sh returnStatus: true, script: 'diff -q -r url1Dir1 url1Dir2'
+                        def excludeFlags=""
+                        def excludeFileList = excludeFiles.split(';')
+                        for(String excludeFile in excludeFileList ) {
```
karianna, line 43:
```suggestion
                        for(String excludeFile in excludeFileList) {
```

karianna (approved):

PR #464, opened 2022-10-13
release: set aqaReference tag for 2022 Oct CPU release

zdtsw:
to set default value of aqaRef to release tag, making sure we only use this for test
**this PR should be reverted once we've done CPU release**

### DIFF
--- a/pipelines/jobs/pipeline_job_template.groovy
+++ b/pipelines/jobs/pipeline_job_template.groovy
@@ -95,7 +95,7 @@ pipelineJob("${BUILD_FOLDER}/${JOB_NAME}") {
         stringParam('buildReference', '', 'Tag name or Branch name of temurin-build repo. Defaults to master')
         stringParam('ciReference', '', 'Tag name or Branch name of ci-jenkins-pipeline repo. Defaults to master')
         stringParam('helperReference', '', 'Tag name or Branch name of jenkins-helper repo. Defaults to master')
-        stringParam('aqaReference', '', 'Tag name or Branch name of aqa-tests. Defaults to master')
+        stringParam('aqaReference', 'v0.9.5-release', 'Tag name or Branch name of aqa-tests. Defaults to master')
         booleanParam('aqaAutoGen', false, 'If set to true, force auto generate AQA test jobs. Defaults to false')
         booleanParam('enableTests', runTests, 'If set to true the test pipeline will be executed')
         booleanParam('enableTestDynamicParallel', runParallel, 'If set to true test will be run parallel')


andrew-m-leonard:
@zdtsw @smlambert So interesting change, so basically once this re-gens, the pipelines nightly/release will use the specific AQA tag, rather than HEAD.. Would we want that ongoing...? ie.by default we use the latest and greatest AQA tag? So if someone wants to run a pipeline on HEAD, they'd need to specify..(blank) ?

zdtsw:
> @zdtsw @smlambert So interesting change, so basically once this re-gens, the pipelines nightly/release will use the specific AQA tag, rather than HEAD.. Would we want that ongoing...? ie.by default we use the latest and greatest AQA tag? So if someone wants to run a pipeline on HEAD, they'd need to specify..(blank) ?

I am thinking for the next week, we can lock to use the aqa release tag, once the build + test in public jenkins is done, we can revert the change, so aqa uses HEAD again.
The reason for this PR is because, i manually set the v0.9.5-release to all opendjkX-pipeline yesterday, expecting over the night we will start to use that one. But we have PRs got merged from outreachy which re-genereate pipelines. so if we do not code freeze next week (even just a *.md change) which can cause our pipeline keeps re-generate. so it runs back and forth between v0.9.5 and HEAD, which need to double check everytime when start a  new release build.

andrew-m-leonard (approved):

karianna:
@zdtsw Eclipse CLA seems to be checking a different user name - is some git metadata getting mixed up here?

zdtsw:
> @zdtsw Eclipse CLA seems to be checking a different user name - is some git metadata getting mixed up here?

Thanks! 
yeah, on a different laptop, not config to use working e-mail.
Will amend it

smlambert:
Makes one wonder whether the automatic regeneration of the build pipelines is a good feature, or if the regen could be tuned to be triggered only when critical pipeline code changes, to avoid some of the unnecessary churn described.  

To have more controls over it and avoid churn, we chose to put a flag on the autogen of test pipelines.

zdtsw:
what we could do is, to manually disable 
- https://ci.adoptopenjdk.net/job/build-scripts/job/utils/job/build-pipeline-generator/
- https://ci.adoptopenjdk.net/job/build-scripts/job/utils/job/pipeline_jobs_generator_jdkXu/ 
after we declare "code freeze" and still allow trivial PRs (documentation)  to be merged to master

the risk of this is:  once we have emergency PR landed, we need to remember enable these jobs and all PRs between will be taken into automatically in that run.

smlambert:
My comment really should go into a release retrospective issue.  I am not opposed to this PR as a way to work around the currently overactive pipeline generation.  I was saying that perhaps after October release we consider making the pipeline generation 'smarter' (as in it won't trigger on doc changes or non-pipeline code changes).

zdtsw:
> My comment really should go into a release retrospective issue. I am not opposed to this PR as a way to work around the currently overactive pipeline generation. I was saying that perhaps after October release we consider making the pipeline generation 'smarter' (as in it won't trigger on doc changes or non-pipeline code changes).

I agree. I will close this PR and put in the retro see how we can improve this for the next CPU.

Issue #51, opened 2021-02-18
sign_build job should be more configurable


AdamBrousseau:
Related/Similar AdoptOpenJDK/ci-jenkins-pipelines#50 

https://github.com/AdoptOpenJDK/openjdk-build/blob/282058709909733381464dcb67b7f0d13dca4e13/pipelines/build/common/openjdk_build_pipeline.groovy#L226

The function for sign build has some hardcoded values which could be parameterized to make the scripts more portable.

#### Lables
https://github.com/AdoptOpenJDK/openjdk-build/blob/282058709909733381464dcb67b7f0d13dca4e13/pipelines/build/common/openjdk_build_pipeline.groovy#L237
`def nodeFilter = "${buildConfig.TARGET_OS}"`

https://github.com/AdoptOpenJDK/openjdk-build/blob/282058709909733381464dcb67b7f0d13dca4e13/pipelines/build/common/openjdk_build_pipeline.groovy#L242
`nodeFilter = "${nodeFilter}&&build&&win2012"`

https://github.com/AdoptOpenJDK/openjdk-build/blob/282058709909733381464dcb67b7f0d13dca4e13/pipelines/build/common/openjdk_build_pipeline.groovy#L248
`nodeFilter = "${nodeFilter}&&macos10.14"`

#### Certificates
https://github.com/AdoptOpenJDK/openjdk-build/blob/282058709909733381464dcb67b7f0d13dca4e13/pipelines/build/common/openjdk_build_pipeline.groovy#L241
`certificate = "C:\\openjdk\\windows.p12"`

https://github.com/AdoptOpenJDK/openjdk-build/blob/282058709909733381464dcb67b7f0d13dca4e13/pipelines/build/common/openjdk_build_pipeline.groovy#L246
`certificate = "\"Developer ID Application: London Jamocha Community CIC\""`

AdamBrousseau:
Not sure why this wasn't handled in #145. This issue is still valid even though the code has changed.

IMO this line should be configurable
https://github.com/adoptium/ci-jenkins-pipelines/blob/c5a5ed7dd940b95951dc3379fb73605c5f31f013/pipelines/build/common/openjdk_build_pipeline.groovy#L404

SehrishHussain:
@andrew-m-leonard can we work on this issue?

SehrishHussain:
[...]

PR #467, opened 2022-10-13
sign node label

cornelia247:
Parameterized the SIGN_NODE_LABEL to allow users conveniently change defaults https://github.com/adoptium/ci-jenkins-pipelines/issues/51

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -485,6 +485,7 @@ class Build {
     Run the Sign downstream job. We run this job on windows and jdk8 hotspot & jdk13 mac builds.
     The job code signs and notarizes the binaries so they can run on these operating systems without encountering issues.
     */
+    def SIGN_NODE_LABEL = DEFAULTS_JSON['signNodeLabel']
     def sign(VersionInfo versionInfo) {
         // Sign and archive jobs if needed
         if (
@@ -493,7 +494,8 @@ class Build {
             context.stage('sign') {
                 def filter = ''
 
-                def nodeFilter = 'eclipse-codesign'
+                def nodeFilter = "${SIGN_NODE_LABEL}"
+                 
 
                 if (buildConfig.TARGET_OS == 'windows') {
                     filter = '**/OpenJDK*_windows_*.zip'
@@ -701,7 +703,8 @@ class Build {
             default: break
         }
 
-        def nodeFilter = 'eclipse-codesign'
+        /* groovylint-disable-next-line VariableTypeRequired */
+        def nodeFilter = "${SIGN_NODE_LABEL}"
 
         // Execute sign installer job
         def installerJob = context.build job: 'build-scripts/release/sign_installer',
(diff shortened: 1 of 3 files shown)


pipelines/defaults.json
```diff
@@ -53,5 +53,6 @@
             "numMachines"    : ["3"]
         }
     },
-    "defaultsUrl"            : "https://raw.githubusercontent.com/adoptium/ci-jenkins-pipelines/master/pipelines/defaults.json"
+    "defaultsUrl"            : "https://raw.githubusercontent.com/adoptium/ci-jenkins-pipelines/master/pipelines/defaults.json",
+    "signNodeLabel"            : "eclipse-codesign"
```
karianna, line 57:
```suggestion
    "signNodeLabel"           : "eclipse-codesign"
```

karianna, line 57: Hmm, I wonder if this should be in Adopt defaults json as opposed to defaults json...

cornelia247: Both default JSONs' variables were mapped and an if statement was created to ensure default json was used if adopt default shell script wasn't in play, I guess it would be safer to use adopt default, I'll make the change and commit now. thanks for the correction

cornelia247: they both seem to be feeding from the same URL, I think it's safe

karianna (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -485,6 +485,7 @@ class Build {
     Run the Sign downstream job. We run this job on windows and jdk8 hotspot & jdk13 mac builds.
     The job code signs and notarizes the binaries so they can run on these operating systems without encountering issues.
     */
+    def SIGN_NODE_LABEL = DEFAULTS_JSON['signNodeLabel']
```
sophia-guo, line 488: It is better to move this definition to the function [build()](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L1533) to be visible to both sign() and  signInstaller()

sophia-guo:
[context.node('eclipse-codesign') ](https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L1304) , feels this node is also used to sign jmods. @andrew-m-leonard is it good to update the line mentioned here to `context.node("${SIGN_NODE_LABEL}")?

zdtsw:
maybe ask [AdamBrousseau](https://github.com/AdamBrousseau) to take a look too, since he is the one created original issue

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -701,7 +703,8 @@ class Build {
             default: break
         }
 
-        def nodeFilter = 'eclipse-codesign'
+        /* groovylint-disable-next-line VariableTypeRequired */
+        def nodeFilter = "${SIGN_NODE_LABEL}"
```
zdtsw, line 707:
```suggestion
        String nodeFilter = "${SIGN_NODE_LABEL}"
```

zdtsw:
@cornelia247 need a rebase first

andrew-m-leonard:
this has been stale for over a year, converting to "draft", if no longer needed please close.

PR #154, opened 2021-05-31
Add job level security to hide non Temurin builds

gdams:
Closes: https://github.com/adoptium/adoptium/issues/6

This change will apply a security matrix to all jobs that are not Temurin builds.

PR #474, opened 2022-10-14
Only force authorizationMatrix on Adoptium's jenkins

### LINKED ISSUE #156: Request to make Jenkins job permissions configurable
https://github.com/adoptium/ci-jenkins-pipelines/pull/154
https://github.com/adoptium/ci-jenkins-pipelines/pull/155

These 2 PRs introduced changes specific to Adopt that do not apply to other consumers of the Jenkins pipelines. Can these be reworked to be configurable via the defaults.json file? Minimum would be a boolean but ideally also moving the hardcoded team name(s) into the config file.

AdamBrousseau:
The changes introduced by #154 #155 are specific
to Adoptium and can cause unwanted behaviour on
other users' Jenkins'. Simple solution to only
run that part of the code on Jenkins' that are
owned by Adopt (have adopt in the URL).

Probably good enough to resolve #156 while we're here.


AdamBrousseau:
https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
>JENKINS_URL
Full URL of Jenkins, such as https://example.com:port/jenkins/ (NOTE: only available if Jenkins URL set in "System Configuration")

Open to suggestions. I figured `adopt` would cover if the server ended up moving to `adoptium` instead of `adoptopenjdk`.

tellison:
I'm sure there will be lots of places to change if we move doman, but I will let others weigh in on the safest way to check where this logic is running.

sxa:
> I'm sure there will be lots of places to change if we move doman, but I will let others weigh in on the safest way to check where this logic is running.

FYI This came out of a [quick brainstorming thread on slack](https://adoptium.slack.com/archives/C53GHCXL4/p1665761567805169) so I'm 👍🏻 on this to resolve the immediate problem. It's better for the default be to not have such a restriction.

karianna (approved):
LGTM

AdamBrousseau:
bump

tellison:
I know that the security of this code is not compromised by allowing this behavior on a mismatched URL, but as stated above it just looks uncomfortable to see code added that doesn't [adhere to good practice](https://codeql.github.com/codeql-query-help/javascript/js-incomplete-url-substring-sanitization/) and I wouldn't want people to use that as an example of how to determine that we are really running on Adoptium infrastructure elsewhere where it does matter. Just my 2c.

AdamBrousseau:
@tellison so preference would be something like
```
JENKINS_URL.contains("ci.adoptopenjdk.net")
```
?

tellison:
> @tellison so preference would be something like
> 
> ```
> JENKINS_URL.contains("ci.adoptopenjdk.net")
> ```
> 
> ?

No, that doesn't work. Look at the link referenced above for reasons.
I feel bad. As I say this code is not validating external data that could be problematic in this situation, so I'm not insisting on this change!

AdamBrousseau:
Sorry I missed the link the first time.
If I understand correctly, an attacker would need to be in control of Adopt's Jenkins' $JENKINS_URL variable in order for the job generation to "expose" OpenJ9 jobs.
Based on the last part of your last comment, are we ok as-is then?

tellison:
> Sorry I missed the link the first time. If I understand correctly, an attacker would need to be in control of Adopt's Jenkins' $JENKINS_URL variable in order for the job generation to "expose"

Yes. As [I wrote](https://github.com/adoptium/ci-jenkins-pipelines/pull/474#issuecomment-1331911216), the proposal is not insecure. It just causes me to twitch when I see code like this which in other contexts would be insecure. Like pointing an unloaded gun at people - it is safe, but you just don't do that :-) 

> OpenJ9 jobs. Based on the last part of your last comment, are we ok as-is then?

I'm not blocking it. It is waiting on reviewers ATM. Apologies for the distraction!

karianna:
@AdamBrousseau Some linter fixes required

AdamBrousseau:
>Some linter fixes required

Resolved

pipelines/build/common/create_job_from_template.groovy
```diff
@@ -61,7 +61,7 @@ pipelineJob("$buildFolder/$JOB_NAME") {
     }
     properties {
         // Hide all non Temurin builds from public view
```
karianna, line 63:
```suggestion
        // Hide all non Temurin builds from public view on the Adoptium CI instance
```

karianna (approved):
LGTM

gdams (approved):

PR #480, opened 2022-10-18
setting sdk_resource as customized for runAqaTest

julian55455:
Fixes https://github.com/adoptium/aqa-tests/issues/2097

cc @sophia-guo @smlambert @llxia


llxia (changes requested):
This is incorrect. The change should be in https://github.com/ibmruntimes/ci-jenkins-pipelines repo, not https://github.com/adoptium/ci-jenkins-pipelines.

The change should be at:
https://github.com/ibmruntimes/ci-jenkins-pipelines/blob/ibm/pipelines/build/common/openjdk_build_pipeline.groovy#L312
https://github.com/ibmruntimes/ci-jenkins-pipelines/blob/ibm/pipelines/build/common/openjdk_build_pipeline.groovy#L446

PR #491, opened 2022-11-09
Fix: comparision job does not take in exclude list and bind to only run x64 machine for now

zdtsw:
- by using '' jenkins does not take in variable excludeFlags so the job still compare classes_nocoops.jsa
- if not set to use x64, currently we might ended on a risvc machine or others(aix?) so we need to test more if wget and tar with currents flags work or not. since this is a small/quick job, it wont hold x64 linux for more than 2mins, should not be a problem. 

Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/435


karianna (approved):
LGTM

andrew-m-leonard (approved):

PR #493, opened 2022-11-14
Add loongson support

Panxuefeng-loongson:
There are no LoongArch machines available for jenkins, but we are actively seeking solutions to the problem, see:https://github.com/adoptium/infrastructure/issues/2806. For now, this changes is only used to seek advice from everyone.

### DIFF
--- a/.github/regex_labeler.yml
+++ b/.github/regex_labeler.yml
@@ -97,6 +97,10 @@ labels:
     matcher:
       title: "bisheng"
       body: "bisheng"
+  - label: "loongson"
+    matcher:
+      title: "loongson"
+      body: "loongson"
   - label: "docker"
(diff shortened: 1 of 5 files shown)


sxa:
ON hold pending https://github.com/adoptium/infrastructure/pull/2839 since we cannot add this if we do not have machines available to run it.

Panxuefeng-loongson:
> ON hold pending [adoptium/infrastructure#2839](https://github.com/adoptium/infrastructure/pull/2839) since we cannot add this if we do not have machines available to run it.

Indeed. So we are very eager to add LoonArch machines to the infrastructure. From your perspective, what else do we need to do for the community to accept LoongArch machines into the infrastructure?

PR #518, opened 2022-11-28
Enable jck extended.jck parallel on xlinux, windows64 and x86-64_macos

sophia-guo:
Relates https://github.com/temurin-compliance/temurin-compliance/issues/268

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -481,20 +481,41 @@ class Build {
         }
         def sdkUrl = "${env.BUILD_URL}/artifact/workspace/target/${filter}/*zip*/target.zip"
         context.echo "sdkUrl is ${sdkUrl}"
-        def targets = 'sanity.jck,extended.jck,special.jck'
-        context.triggerRemoteJob abortTriggeredJob: true,
-                            blockBuildUntilComplete: false,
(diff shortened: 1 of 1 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -481,20 +481,41 @@ class Build {
...
+        def parallel = 'None'
+        def num_machines = '1'
+        def remoteTargets = [:]
+
+        targets.each { target ->
+            // For each requested test, i.e 'sanity.jck', 'extended.jck', 'special.jck', call test job
+            try {
+                context.println "Remote trigger ${target}"
+                remoteTargets["${target}"] = {
+                    if ( "${target}" == 'extended.jck' && ("${platforms}" == 'x86-64_linux' || "${platforms}" == 'x86-64_windows')) {
```
sxa, line 494: As discussed yesterday I suggest adding `x86-64_macos` as it is a primary and we can generally aim to run this over two machines. Worst case a machien will be offline, but we should aim to run on all available machines since that platform takes over 24 hours to complete.

sxa (changes requested):

pipelines/build/common/openjdk_build_pipeline.groovy
sophia-guo: Enabled with x86-64_macos

sxa (approved):

Issue #514, opened 2022-11-24
Support generation of "release" pipeline and build jobs for "release build Tag" support


andrew-m-leonard:
The new "build ref tag" support for releases needs an extension to the generation build job to support generating a "Release" pipeline and build jobs based off the given "build-ref's" (temurin-build, ci-jenkins-pipeline, helper). So you would manually run the generation jobs with an option "Build-ref tag: xxx". Which then generates, eg:

- release-openjdk17-pipeline
- release-jobs/jdk17u/jdk17u-aix-ppc-..., .....

PR #525, opened 2022-12-02
update: remove aarch32 jdk8 from release-pipeline default setting

zdtsw:
some more for the release pipeline
	- do not have aarch32 in the same batch of build by release-openjdk8-pipeline due to different scmReference
	- good to manually trigger that release by sharing the same pipeline with other jdk8 targets
	- make additionalConfigureArgs only set for jdk11+, default to ' '

Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/514

### DIFF
--- a/pipelines/build/regeneration/release_pipeline_generator.groovy
+++ b/pipelines/build/regeneration/release_pipeline_generator.groovy
@@ -90,7 +90,7 @@ node('worker') {
                 config.put('defaultsJson', DEFAULTS_JSON)
                 config.put('adoptDefaultsJson', ADOPT_DEFAULTS_JSON)
 
-                if(${javaVersion} != "8") { // for jdk11+, need extra config args to pass down
+                if(${javaVersion} >= "11") { // for jdk11+, need extra config args to pass down
                     config.put('additionalConfigureArgs', "--without-version-pre --without-version-opt")
                 }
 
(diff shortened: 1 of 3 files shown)


pipelines/build/regeneration/release_pipeline_generator.groovy
```diff
@@ -90,6 +90,10 @@ node('worker') {
                 config.put('defaultsJson', DEFAULTS_JSON)
                 config.put('adoptDefaultsJson', ADOPT_DEFAULTS_JSON)
 
+                if(${javaVersion} != "8") { // for jdk11+, need extra config args to pass down
```
karianna, line 93: Maybe that should be >= 11?

karianna (changes requested):
Are the docs updated to reflect how to run this arm32 Java 8 build separately during a release cycle?

zdtsw:
> Are the docs updated to reflect how to run this arm32 Java 8 build separately during a release cycle?

I have a draft PR https://github.com/adoptium/temurin-build/pull/3159/files for all the changes we make for next release. 
Still more things to be added for "auto" trigger and some.

karianna (approved):
LGTM assuming the arm32 docs/instructions get landed at the same time

zdtsw:
> LGTM assuming the arm32 docs/instructions get landed at the same time

docs will be merged before 2023 Jan release as new instruction of how to generate/use/trigger release pipelines.

andrew-m-leonard (approved):

PR #537, opened 2022-12-13
Ensure we do not publish fast_startup artifacts

andrew-m-leonard:
Fixes failing jdk11u nightly publishing, where the release script publish with the updated publishing validation reports finding fast_startup variant tarballs that should not be published.


andrew-m-leonard:
Going to change publish script instead

PR #539, opened 2022-12-14
Trigger remote jck with just the jdkUrl

andrew-m-leonard:
The Temurin remote jck trigger was zipping all the artifacts creating a very large zip, when just the jdk is required, which hope might help avoid glitches on the other end during curl downloads..

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -475,9 +475,10 @@ class Build {
     def remoteTriggerJckTests(String platform) {
         def jdkVersion = getJavaVersionNumber()
         //def sdkUrl="https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk${jdkVersion}-pipeline/${env.BUILD_NUMBER}/"
-        def filter = "*.tar.gz"
+        // We just need the JDK for Jck tests
+        def filter = "*-jdk_*.tar.gz"
         if (buildConfig.TARGET_OS.contains("windows")) {
-        	filter = "*.zip"
+        	filter = "*-jdk_*.zip"
         }
         def sdkUrl = "${env.BUILD_URL}/artifact/workspace/target/${filter}/*zip*/target.zip"
         context.echo "sdkUrl is ${sdkUrl}"


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -474,13 +474,14 @@ class Build {
     // Temurin remote jck trigger
     def remoteTriggerJckTests(String platform) {
         def jdkVersion = getJavaVersionNumber()
-        //def sdkUrl="https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk${jdkVersion}-pipeline/${env.BUILD_NUMBER}/"
-        def filter = "*.tar.gz"
+        def jdkUrl = "${env.BUILD_URL}/artifact/workspace/target/${buildConfig.TARGET_OS}/${buildConfig.ARCHITECTURE}/${buildConfig.VARIANT}/*-jdk_*"
```
sophia-guo, line 477: Is this URL correct? I feels like this is the url for pipeline build not jdk build?

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -474,13 +474,14 @@ class Build {
     // Temurin remote jck trigger
     def remoteTriggerJckTests(String platform) {
         def jdkVersion = getJavaVersionNumber()
-        //def sdkUrl="https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk${jdkVersion}-pipeline/${env.BUILD_NUMBER}/"
-        def filter = "*.tar.gz"
```
sophia-guo, line 478: This could be done by update `def filter = "*-jdk_*.tar.gz" ` and for windows ` filter = "*-jdk_*.zip"`

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: thank you @sophia-guo you're correct

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: thanks again, yeah I got to here in a roundabout route !

andrew-m-leonard: used you suggestion

andrew-m-leonard:
@sophia-guo thanks for the comment, should be good now
i've just running a linux and a windows test now

sophia-guo (approved):

Issue #538, opened 2022-12-14
jdk20 pipeline is broken


zdtsw:
start to see problem since https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk20-pipeline/107/console
on "nightly" build,  it cannot find valid `testList`
`java.lang.NullPointerException: Cannot invoke method unique() on null object`

suspect: https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/jobs/configurations/jdk20_pipeline_config.groovy#L167 is not supported
need an empty `"night:[]"`
https://github.com/adoptium/ci-jenkins-pipelines/pull/524

sophia-guo:
Yes, if the testList is customized both nightly and weekly are required.  Otherwise it's unclear if it means default (defined by default.jason) or it's empty list.

PR #542, opened 2022-12-15
fix: broken jdk20 pipeline when missing nightly as entity

zdtsw:
Fix: https://github.com/adoptium/ci-jenkins-pipelines/issues/538


andrew-m-leonard (approved):
lgtm

sophia-guo (approved):

Issue #473, opened 2022-10-14
Introduce a new "prototype" pipeline for new potential platforms to initially run in


andrew-m-leonard:
New platforms that are not candidates for release need time to bed in and prove their stability. We should introduce a new "prototype" (or suitably named) pipeline that is perhaps run once a week for such platforms.

andrew-m-leonard:
hi @cornelia247 , this item is quite extensive, but we will not be implementing it until after the October release, so into November. Also, this requires certain Admin access for some of the updates/testing.

zdtsw:
I think Loongson could be a good candidate for this issue.

zdtsw:
weekly build for jdk11 has bisheng corretto and dragonwell, this prevents archive temurin.
We should consider moving these 3 variants into "prototype"

zdtsw:
^ this is also on jdk17 https://ci.adoptopenjdk.net/job/build-scripts/job/weekly-openjdk17-pipeline/: bisheng and openj9

zdtsw:
maybe risc-v should also be in the prototype, rather than manually created by coping  from old jobs

zdtsw:
Mark: when code is done, we need to update docs of "how to generate these prototype pipelines"

zdtsw:
one more question: do we want to run the targets of prototype in the " weekly"? @andrew-m-leonard 
I am thinking, we can have the same scheduler (as-is) for both "nightly" and "prototype" two to three times a week, but weekly only runs for nightlies' 
so we probably can get weekly's result of AQA tests within 24hrs (will be much faster than now, when we move non-temurin variants into prototype)

andrew-m-leonard:
> one more question: do we want to run the targets of prototype in the " weekly"? @andrew-m-leonard I am thinking, we can have the same scheduler (as-is) for both "nightly" and "prototype" two to three times a week, but weekly only runs for nightlies' **=> we do not change logic in the weekly->nightly** so we probably can get weekly's result of AQA tests within 24hrs (will be much faster than now, when we move non-temurin variants into prototype)

@zdtsw Yes agree, weekly just runs the nightly targets

sxa:
Just to be clear, are we suggesting in the last two comments that the prototype builds would NOT be included in the weekly runs? That would seem undesirable as it would not allow us to regularly run and see the output of the full set of test suites on the prototype platforms.

zdtsw:
We will not have prototype running in the "normal" weekly pipeline,
[...]

PR #545, opened 2022-12-16
Switch testing for riscv64Linux to default

luhenry:
It's passing all nightly builds in ~15hrs already, so we should be good to run it nightly. Also, it would be very worthwhile to get more test coverages, similarly to other platforms.


sxa (approved):
LGTM and we can get this running as part of https://github.com/adoptium/ci-jenkins-pipelines/issues/473

zdtsw (approved):

PR #555, opened 2023-01-02
update: enable evaluation by set timer

zdtsw:
Ref: #473

### DIFF
--- a/pipelines/jobs/configurations/jdk11u.groovy
+++ b/pipelines/jobs/configurations/jdk11u.groovy
@@ -1,7 +1,6 @@
 targetConfigurations = [
         'x64Mac'        : [    'temurin',    'openj9'                    ],
         'x64Linux'      : [    'temurin',    'openj9',    'dragonwell',    'corretto',    'bisheng',    'fast_startup'],
-        'x64AlpineLinux': [    'temurin'                            ],
         'x64Windows'    : [    'temurin',    'openj9',    'dragonwell'            ],
         'x32Windows'    : [    'temurin'                            ],
         'ppc64Aix'      : [    'temurin',    'openj9'                    ],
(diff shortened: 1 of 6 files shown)


pipelines/jobs/configurations/jdk11u.groovy
```diff
@@ -1,7 +1,6 @@
 targetConfigurations = [
         'x64Mac'        : [    'temurin',    'openj9'                    ],
         'x64Linux'      : [    'temurin',    'openj9',    'dragonwell',    'corretto',    'bisheng',    'fast_startup'],
-        'x64AlpineLinux': [    'temurin'                            ],
```
gdams, line 4: wrong platform? This PR says it moves the aarch64 alpine build?

zdtsw: yes, my bad, will fix it

pipelines/jobs/configurations/jdk11u_evaluation.groovy
```diff
@@ -1,4 +1,10 @@
 targetConfigurations = [
+        'aarch64Linux'  : [
```
gdams, line 2: still not Alpine Linux? perhaps I'm missing something here?

zdtsw: see it as my covid side effects.......

we actually do not have jdk11 testing  aarch64alpinelinux.
will not have it in evaluation.

revert back

zdtsw: there was some issue for jdk11 aarch64 alpine linux in the past, so it was disabled on jdk11
will first get other jdk version running and put it into evaluation later.

andrew-m-leonard (approved):

karianna (approved):
LGTM

PR #566, opened 2023-01-06
Add AIX to jdk17&19 release configuration

andrew-m-leonard:
AIX was missing from the jdk-17 and jdk-19 release pipeline configuration and hence was not being built.

### DIFF
--- a/pipelines/jobs/configurations/jdk17u_release.groovy
+++ b/pipelines/jobs/configurations/jdk17u_release.groovy
@@ -17,6 +17,9 @@ targetConfigurations = [
         'ppc64leLinux': [
                 'temurin'
         ],
+        'ppc64Aix'    : [
+                'temurin'
+        ],
         's390xLinux'  : [
                 'temurin'
         ],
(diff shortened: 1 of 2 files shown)


pipelines/jobs/configurations/jdk17u_release.groovy
```diff
@@ -17,6 +17,9 @@ targetConfigurations = [
         'ppc64leLinux': [
                 'temurin'
         ],
+        'ppc64Aix'    : [
```
zdtsw, line 20: should we build AIX for jdk17 /19? if we do not officially release it out

zdtsw: otherwise i am happy to approve this

smlambert: We want to officially release it, but had an issue in Temurin-compliance that was blocking it (which was why it was not added originally).  Think Stewart & Tyler have arrived at a solution to the blocking issue, so we can reconstitute it.

smlambert (approved):

zdtsw (approved):

PR #569, opened 2023-01-10
update: weekly evaluation need to set to Weekly not Nightly to run aqa weekly test

zdtsw:
need overwrite releaseType for weekly-evaluation, to be able to run Aqa weekly tests but not trigger TCK

Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/473


steelhead31 (approved):
LGTM!

Issue #571, opened 2023-01-13
Release pipeline generation make use of "USE_ADOPT_SHELL_SCRIPTS": false, which fails on Mac


andrew-m-leonard:
Release pipeline jobs use the "user" configuration pointing at the user remote tag/branch, but the necessary dual pass Mac eclipse-codesign depends on  "USE_ADOPT_SHELL_SCRIPTS": true.

zdtsw:
so to not rush for a big change just before 2023 Jan release, propose to make a workaround for Mac release pipeline.
the idea is to: passing down USE_ADOPT_SHELL_SCRIPTS when `release-openjdkXX-pipeline` trigger downstream job 
`jdkXX u-release-mac-x64-temurin` and `jdkXXu-release-mac-aarch64-temurin`

so the jobs `jdkXX u-release-mac-x64-temurin` and `jdkXXu-release-mac-aarch64-temurin` will still use "master" branches from ci-jenkins-pipeline and temurin-build for this release.



We still need a proper long term solution for this fix.

zdtsw:
some thoughts for future fix:
1.  less code change but add new step before release process
before we create release tag (e.g v2023.01.09) for ci-jenkins-pipeline and temurin-build and helper, we create a new commit to change the content of defaults.json (snippet and assume two new tags both called v2023.01.09)
from 
```
    "repository"             : {
        "build_url"          : "https://github.com/adoptium/temurin-build.git",
        "build_branch"       : "master",
        "test_dirs"          : "/test/functional",
        "pipeline_url"       : "https://github.com/adoptium/ci-jenkins-pipelines.git",
        "pipeline_branch"    : "master",
        "helper_ref"         : "master"
    },
```

to 
```
    "repository"             : {
        "build_url"          : "https://github.com/adoptium/temurin-build.git",
        "build_branch"       : "v2023.01.09",
        "test_dirs"          : "/test/functional",
        "pipeline_url"       : "https://github.com/adoptium/ci-jenkins-pipelines.git",
        "pipeline_branch"    : "v2023.01.09",
        "helper_ref"         : "v2023.01.09"
    },

```
along with changes from https://github.com/zdtsw/ci-jenkins-pipelines/tree/issue/571_2

in this case, we have USE_ADOPT_SHELL_SCRIPTS to true for all release pipelines and jobs

[...]

PR #572, opened 2023-01-13
workaround: to use USE_ADOPT_SHELL_SCRIPTS for Mac build

zdtsw:
- this is a workaround to make release work on mac
	- code should be reverted when we have a formal solution
Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/571


andrew-m-leonard (approved):
ok for workaround

steelhead31 (approved):
Looks ok.

PR #575, opened 2023-01-14
fix: override USE_ADOPT_SHELL_SCRIPTS to true for mac in release pipeline

zdtsw:
fix error:
`groovy.lang.ReadOnlyPropertyException: Cannot set readonly property: USE_ADOPT_SHELL_SCRIPTS for class: common.IndividualBuildConfig`

ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/571 workaround


zdtsw:
this wont work, as the same error to overwrite java class attribute. 
solution has been changed to update defaults.json instead of manipulating attribute

PR #576, opened 2023-01-14
fix: to enable USE_ADOPT_SHELL_SCRIPTS with defaults.json in release pipeline 

zdtsw:
set defaults.json to use tag v2023.01.03

also revert the workaround done in previous commit

Ref: https://github.com/adoptium/ci-jenkins-pipelines/issues/571

### DIFF
--- a/pipelines/build/regeneration/build_job_generator.groovy
+++ b/pipelines/build/regeneration/build_job_generator.groovy
@@ -27,6 +27,9 @@ limitations under the License.
 
 String javaVersion = params.JAVA_VERSION
 String ADOPT_DEFAULTS_FILE_URL = 'https://raw.githubusercontent.com/adoptium/ci-jenkins-pipelines/master/pipelines/defaults.json'
+if (params.REPOSITORY_BRANCH != "") { // mainly for release jobs to use tag
+    ADOPT_DEFAULTS_FILE_URL = "https://raw.githubusercontent.com/adoptium/ci-jenkins-pipelines/${params.REPOSITORY_BRANCH}/pipelines/defaults.json"
+}
 String DEFAULTS_FILE_URL = (params.DEFAULTS_URL) ?: ADOPT_DEFAULTS_FILE_URL
 
 node('worker') {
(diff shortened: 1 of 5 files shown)


zdtsw:
to proceed with this commit for release pipeline, 
we need to remove the v2023.01.03 tag from origin and retag it on this commit (as merge to master) than rerun the pipeline re-generator.

andrew-m-leonard:
@zdtsw Thank you for looking at this.
Should we create a branch instead?, on each repo, say called "v2023.01", then merge this into that branch ?

zdtsw:
> @zdtsw Thank you for looking at this. Should we create a branch instead?, on each repo, say called "v2023.01", then merge this into that branch ?

we can do that way as well. just need to be careful, a branch in origin, cannot rebase or squash etc. that will impact any one clone origin. otherwise i think it should work.

andrew-m-leonard:
> so someone need to create the v2023.01 branch in origin then i can change this PR to this new branch instead of "master" but the part of the revert previous commit is still needed to clean it out form "master",

@zdtsw branch v2023.01 created for all 3 repos: ci-jenkins-pipelines, temurin-build and jenkins-helper

pipelines/build/regeneration/build_job_generator.groovy
```diff
@@ -27,6 +27,9 @@ limitations under the License.
 
 String javaVersion = params.JAVA_VERSION
 String ADOPT_DEFAULTS_FILE_URL = 'https://raw.githubusercontent.com/adoptium/ci-jenkins-pipelines/master/pipelines/defaults.json'
+if (params.REPOSITORY_BRANCH != "") { // mainly for release jobs to use tag
+    ADOPT_DEFAULTS_FILE_URL = "https://raw.githubusercontent.com/adoptium/ci-jenkins-pipelines/${params.REPOSITORY_BRANCH}/pipelines/defaults.json"
```
andrew-m-leonard, line 31: Do we need this if this code is on a branch? ie. line 29 can just be set to /v2023.01/....

zdtsw: yes, otherwise it will still use "master" line 29's content.
and it would be good, this if--block can be eventually apply to "master" not only live on v2023.01

PR #587, opened 2023-01-16
Pass smokeTest parameter VENDOR_TEST_BRANCHES as variable

### LINKED ISSUE #585: Be able to re-generate smoketest jobs for release pipeline
When we have a new release tag or branch to re-generate release pipeline and jobs, smoke test aint getting updated.
Because of the logic in the current implementation, smoketest job only get generated if it was not there/or the job is not runnable.

Detail see: https://github.com/adoptium/temurin-build/issues/3212#issuecomment-1382827921

sophia-guo:
Related #https://github.com/adoptium/temurin-build/issues/3212

Close #585 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -148,15 +148,12 @@ class Build {
         jobParams.put('BUILD_LIST', 'functional/buildAndPackage')
         def useAdoptShellScripts = Boolean.valueOf(buildConfig.USE_ADOPT_SHELL_SCRIPTS)
         def vendorTestRepos = ((String)ADOPT_DEFAULTS_JSON['repository']['build_url']) - ('.git')
-        def vendorTestBranches = ADOPT_DEFAULTS_JSON['repository']['build_branch']
         def vendorTestDirs = ADOPT_DEFAULTS_JSON['repository']['test_dirs']
         if (!useAdoptShellScripts) {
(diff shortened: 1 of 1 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -281,7 +278,12 @@ class Build {
     */
     def runSmokeTests() {
         def additionalTestLabel = buildConfig.ADDITIONAL_TEST_LABEL
-
+        def useAdoptShellScripts = Boolean.valueOf(buildConfig.USE_ADOPT_SHELL_SCRIPTS)
+        def vendorTestBranches = ADOPT_DEFAULTS_JSON['repository']['build_branch']
```
zdtsw, line 282: it is safe to use `buildConfig.build_branch`

sophia-guo:
Use Adotp_defaults_Json if USE_ADOPT_SHELL_SCRIPTS is set. Otherwise use Default_JSON. Currently the release pipeline would like to set both adopt_defaults_json and default_json with build tags. However the  build jdk job only default_json is set correctly https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/release/job/jobs/job/jdk17u/job/jdk17u-release-linux-x64-temurin/configure , @zdtsw will work to update that.

smlambert (approved):

zdtsw:
related to fix in  https://github.com/adoptium/ci-jenkins-pipelines/pull/592

Issue #590, opened 2023-01-17
Checksum file may be missing


Accelerator1996:
The steps to generate checksum files are defined in each sign() method, but these are controlled by options similar to enableSigner. Logically speaking, the generation of sha256.txt should not be controlled by these, but should always be generated

zdtsw:
If I understand you correctly, you are trying to generate `*.sha256.txt` as a default behavior even when `enableSigner` is not set to `true`

Accelerator1996:
> If I understand you correctly, you are trying to generate `*.sha256.txt` as a default behavior even when `enableSigner` is not set to `true`

Yes, you are right. I think each of them should have a verification file.

andrew-m-leonard:
@Accelerator1996 Can you point me to the scenario when no sha256.text is generated please? Do you have an example? As the base pipeline should do that:
https://github.com/adoptium/ci-jenkins-pipelines/blob/6b93fad7b37df5f2a50c6740c419b65d8f70b891/pipelines/build/common/build_base_file.groovy#L855

Accelerator1996:
> @Accelerator1996 Can you point me to the scenario when no sha256.text is generated please? Do you have an example? As the base pipeline should do that:
> 
> https://github.com/adoptium/ci-jenkins-pipelines/blob/6b93fad7b37df5f2a50c6740c419b65d8f70b891/pipelines/build/common/build_base_file.groovy#L855

We use kick_off_build.groovy to build our jdk instead of openjdk_pipeline.groovy. Please take a look at this PR [link](https://github.com/adoptium/ci-jenkins-pipelines/pull/591/files). In the file pipelines/build/common/openjdk_build_pipeline.groovy, if enbaleSigner and enableInstallers are False, the checksum files will not be generated, but I think he should be able to generate it in any case.

andrew-m-leonard:
> > @Accelerator1996 Can you point me to the scenario when no sha256.text is generated please? Do you have an example? As the base pipeline should do that:
> > https://github.com/adoptium/ci-jenkins-pipelines/blob/6b93fad7b37df5f2a50c6740c419b65d8f70b891/pipelines/build/common/build_base_file.groovy#L855
> 
[...]

PR #591, opened 2023-01-17
Fix a bug about missing checksum file(#590)

Accelerator1996:
closes: #590 

The version data is parsed in each build, and the files are differentiated by file type. So it is redundant to generate sha256.txt for individual suffix files in sign method

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -613,8 +613,6 @@ class Build {
                             target: 'workspace/target/',
                             flatten: true)
 
-                    context.sh 'for file in $(ls workspace/target/*.tar.gz workspace/target/*.zip); do sha256sum "$file" > $file.sha256.txt ; done'
-
                     writeMetadata(versionInfo, false)
                     context.archiveArtifacts artifacts: 'workspace/target/*'
                 }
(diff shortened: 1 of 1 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -739,7 +737,6 @@ class Build {
                 // Archive the Mac and Windows pkg/msi
                 if (buildConfig.TARGET_OS == 'mac' || buildConfig.TARGET_OS == 'windows') {
                     try {
-                        context.sh 'for file in $(ls workspace/target/*.tar.gz workspace/target/*.pkg workspace/target/*.msi); do sha256sum "$file" > $file.sha256.txt ; done'
```
zdtsw, line 742: By removing this part, installers for window and mac, wont get a *.sha256.txt file.
same as line 768

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1140,6 +1136,16 @@ class Build {
             data.binary_type = type
             data.sha256 = hash
 
+            // generate sha256.txt
```
zdtsw, line 1139: this block is similar to the logic in line 1126, except  skip sbom*
why not just reuse $hash and $file

andrew-m-leonard:
@Accelerator1996 Yes, sha256.txt files produced, : https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk19-pipeline/293/artifact/target/windows/x64/temurin/

PR #592, opened 2023-01-17
fix: pass down ADOPT_DEFAULT_JSON when create downstream job

zdtsw:
- the content might be different if the generator job use a different branch than master

### DIFF
--- a/pipelines/build/common/config_regeneration.groovy
+++ b/pipelines/build/common/config_regeneration.groovy
@@ -30,6 +30,7 @@ class Regeneration implements Serializable {
     private final Map<String, Map<String, ?>> buildConfigurations
     private final Map<String, ?> targetConfigurations
     private final Map<String, ?> DEFAULTS_JSON
+    private final Map<String, ?> ADOPT_DEFAULTS_JSON
     private final Map<String, ?> excludedBuilds
     private Integer sleepTime
     private final currentBuild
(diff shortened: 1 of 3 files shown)


sophia-guo:
Class Regeneration will also need updated? https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/config_regeneration.groovy#L59-L97?

zdtsw:
run tests quick 19

zdtsw:
run tests quick 19

zdtsw:
run tests quick 19

andrew-m-leonard (approved):
lgtm

sophia-guo:
I think to make this PR work  ADOPT_DEFAULTS_JSON need to be reset https://github.com/adoptium/ci-jenkins-pipelines/blob/v2023.01/pipelines/build/regeneration/build_job_generator.groovy#L37-L38 ( which will make the ADOPT_DEFAULTS_JSON to be the default master one) @zdtsw

andrew-m-leonard:
> I think to make this PR work ADOPT_DEFAULTS_JSON need to be reset https://github.com/adoptium/ci-jenkins-pipelines/blob/v2023.01/pipelines/build/regeneration/build_job_generator.groovy#L37-L38 ( which will make the ADOPT_DEFAULTS_JSON to be the default master one) @zdtsw

@sophia-guo @smlambert @zdtsw So this code in this PR has not been run as we haven't run the re-gen.., so I think we keep it that way, thereby we know the build jobs should work as they did for the dry-run.
Yes, I think that means we're building with build-scripts "master" branch not the v2023.01, but I think it's too late to try to fix..
```
[Pipeline] echo
17:28:50  [CHECKOUT] Checking out Adopt Build https://github.com/adoptium/temurin-build.git : master
[Pipeline] checkout
```
Correct me if there's something I might have missed?

sophia-guo:
The Pr is expected to set ADOPT_DEFAULT_JSON  same as DEFAULT_JSON for build jobs, which is related with https://github.com/adoptium/ci-jenkins-pipelines/pull/587. 

Agreed it's not necessary to be in this release as current there is no issue with smoke tests. Can simply update to master https://github.com/adoptium/ci-jenkins-pipelines/pull/595

PR #599, opened 2023-01-31
Add script to show reproducibility of Linux releases

sxa:
As discussed with Andrew - this script is very terse but successfully demonstrates that a build can be reproduced.
I had an initial issue that was related to not having `/usr/local/gcc*/bin` in the path that caused the system `binutils` to be used that broke reproducibility, but that is fixed in this version. Also note the explicit `/usr/bin` in the PATH to ensure that `/bin` isn't used since that makes the logs during `configure` more consistent with the real build process.

This is a good basis on which to provide an end-user build reproducibility comparison too.

### DIFF
--- a/tools/reproduce_comparison/compareLinux.sh
+++ b/tools/reproduce_comparison/compareLinux.sh
@@ -1,0 +1,62 @@
+#!/bin/sh
+[ $# -lt 1 ] && echo Usage: $0 SBOMURL && exit 1
+SBOMURL=$1
+if test -r /etc/redhat-release; then
+  yum install -y gcc gcc-c++ make autoconf unzip zip alsa-lib-devel cups-devel   libXtst-devel libXt-devel libXrender-devel libXrandr-devel libXi-devel
+  yum install -y file fontconfig fontconfig-devel systemtap-sdt-devel # Not included above ...
+  yum install -y git bzip2 xz openssl pigz which # pigz/which not strictly needed but help in final compression
+  if grep elease.6 /etc/redhat-release; then
+    if [ ! -r /usr/local/bin/autoconf ]; then
+      curl https://ftp.gnu.org/gnu/autoconf/autoconf-2.69.tar.gz | tar xpfz - || exit 1
+      (cd autoconf-2.69 && ./configure --prefix=/usr/local && make install)
+    fi
+  fi
+fi
+# ant required for --create-sbom
+if [ ! -r /usr/local/apache-ant-1.10.5/bin/ant ]; then
+  echo Downloading ant for SBOM creation:
+  curl https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.5-bin.zip > /tmp/apache-ant-1.10.5-bin.zip
+  (cd /usr/local && unzip -qn /tmp/apache-ant-1.10.5-bin.zip)
+  rm /tmp/apache-ant-1.10.5-bin.zip
+fi
+echo Retrieving and parsing SBOM from "$SBOMURL"
+curl -LO $SBOMURL
+SBOM=`basename $SBOMURL`
+BOOTJDK_VERSION=$(grep configure_arguments $SBOM | tr ' ' \\n | grep ^Temurin- | uniq | cut -d- -f2)
+GCCVERSION=$(cat $SBOM | tr ' ' \\n | grep CC= | cut -d- -f2 | cut -d\\ -f1)
+LOCALGCCDIR=/usr/local/gcc$(echo $GCCVERSION | cut -d. -f1)
+TEMURIN_BUILD_SHA=$(awk -F'"' '/buildRef/{print$4}' $SBOM  | cut -d/ -f7)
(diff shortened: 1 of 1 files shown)


tools/reproduce_comparison/compareLinux.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+[ $# -lt 1 ] && echo Usage: $0 SBOMURL && exit 1
```
karianna, line 2: Do we need a set -ev?

tools/reproduce_comparison/compareLinux.sh
```diff
@@ -0,0 +1,62 @@
...
+SBOMURL=$1
+#apt install gcc     g++ make autoconf file unzip zip libasound2-dev libcups2-dev libx11-dev libxext-dev libxrender-dev libxrandr-dev libxtst-dev libxt-dev
```
karianna, line 4: commented out code

tools/reproduce_comparison/compareLinux.sh
```diff
@@ -0,0 +1,62 @@
...
+if test -r /etc/redhat-release; then
```
karianna, line 5: two if blocks could be turned into functions?

tools/reproduce_comparison/compareLinux.sh
```diff
@@ -0,0 +1,62 @@
...
+  yum install -y gcc gcc-c++ make autoconf unzip zip alsa-lib-devel cups-devel   libXtst-devel libXt-devel libXrender-devel libXrandr-devel libXi-devel
+  yum install -y file fontconfig fontconfig-devel systemtap-sdt-devel # Not included above ...
+  yum install -y git bzip2 xz openssl pigz which # pigz/which not strictly needed but help in final compression
+  if grep elease.6 /etc/redhat-release; then
+    curl https://ftp.gnu.org/gnu/autoconf/autoconf-2.69.tar.gz | tar xpfz - || exit 1
+    (cd autoconf-2.69 && ./configure --prefix=/usr/local && make install)
+  fi
+fi
+# ant required for --create-sbom
+if [ ! -r /usr/local/apache-ant-1.10.5/bin/ant ]; then
```
karianna, line 15: extract version strings to top of file

tools/reproduce_comparison/compareLinux.sh
```diff
@@ -0,0 +1,62 @@
...
+  rm /tmp/apache-ant-1.10.5-bin.zip
+fi
+echo Retrieving and parsing SBOM from "$SBOMURL"
+curl -LO $SBOMURL
+SBOM=`basename $SBOMURL`
+BOOTJDK_VERSION=$(grep configure_arguments $SBOM | tr ' ' \\n | grep ^Temurin- | uniq | cut -d- -f2)
+GCCVERSION=$(cat $SBOM | tr ' ' \\n | grep CC= | cut -d- -f2 | cut -d\\ -f1)
+LOCALGCCDIR=/usr/local/gcc$(echo $GCCVERSION | cut -d. -f1)
+TEMURIN_BUILD_SHA=$(awk -F'"' '/buildRef/{print$4}' $SBOM  | cut -d/ -f7)
+#TEMURIN_BUILD_ARGS=$(grep makejdk_any_platform_args $SBOM | cut -d\" -f4 | sed -e "s/--disable-warnings-as-errors --enable-dtrace/'--disable-warnings-as-errors --enable-dtrace'/" -e 's/\\n//g' -e "s,/usr/lib/jvm/jdk-[^ ]*,/usr/lib/jvm/jdk-$BOOTJDK_VERSION,g")
```
karianna, line 28: commented out code

karianna (changes requested):
named functions for each discrete chunk will help here :-)

sxa:
@karianna as mentioned in the description this is very much a proof-of-concept at the moment that I wanted to get under version control for others to be able to play with instead of being a polished article. I'm absolutely expecting lots of changes to make it a bit more sane going forwards :-) There is a LOT of things that would need to be changed for this to be a production ready script in addition to some of the ones that you've mentioned.

So I'd prefer to merge this as-is and add the niceties (is that how you spell it?) later, including more sensible parsing of the SBOM.

tools/reproduce_comparison/compareLinux.sh
sxa: Extracted out `ANT_VERSION`

tools/reproduce_comparison/compareLinux.sh
sxa: > Do we need a set -ev?

With the construct in there it's exiting appropriately when the test is false so I do not believe so, but feel free to try and convince me otherwise.

tools/reproduce_comparison/compareLinux.sh
sxa: I've done this, although I honestly don't think it really aids the readability of a script of this size. Also it restricts which shell it can run on since the first line is currently `/bin/sh`  (Admittedly not a problem on RH/CentOS Linux though, which is what script is designed for!)

karianna (approved):

sxa:
Pushed a squash of some of the commits (no changes to code)
@andrew-m-leonard Are you happy to review this? Would prefer the two commits to be in here when it's ready so please use rebase and merge when approved, or leave me to do it after approving :-)

andrew-m-leonard (approved):
yep, thanks

PR #605, opened 2023-02-03
Remote trigger with SETUP_JCK_RUN parameter

sophia-guo:
Release SETUP_JCK_RUN=true, weekly or evaluation weekly SETUP_JCK_RUN=false


sophia-guo:
Related or depends on https://github.com/adoptium/aqa-tests/pull/4239

karianna (approved):
LGTM

karianna:
@sophia-guo Still needed?

sophia-guo:
This is to limit setup_jck_run to release only. If we'd like to also keep it to weekly run the PR will not need any more. @smlambert idea?

smlambert:
Release only is sufficient, as we will also get it when doing a 'dry-run' release pipeline.  Thanks @sophia-guo !

PR #628, opened 2023-02-27
WIP: Retire nightly jdk19 builds and replace with jdk20

adamfarley:
Also includes a few formatting changes for tidiness.


adamfarley:
@steelhead31 - Heya Scott. I think it's a good idea to add a requirement for this sort of PR into the release checklist as a post-release event, to cover releases that are a quarterly CPU containing a short-term release with no subsequent CPUs (e.g. the January release in regard to jdk19).

What do you think?

steelhead31:
> @steelhead31 - Heya Scott. I think it's a good idea to add a requirement for this sort of PR into the release checklist as a post-release event, to cover releases that are a quarterly CPU containing a short-term release with no subsequent CPUs (e.g. the January release in regard to jdk19).
> 
> What do you think?

Definitely a good idea, I've got a number of changes to make to the release checklist, which Im planning to do tomorrow, alongside a number of other documentation changes following the last CPU, so I'll be sure to include this!

gdams (approved):

PR #649, opened 2023-03-08
Propagate downstream build & test results for pipeline.

andrew-m-leonard:
Changes the pipeline propagateFailures param to default to "True".
Also, ensure all sub-jobs including tests are propagated.

Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/234

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -828,13 +828,15 @@ class Builder implements Serializable {
                             // Pass down DEFAULTS_JSON
                             buildJobParams.add(['$class': 'TextParameterValue', name: 'DEFAULTS_JSON', value: JsonOutput.prettyPrint(JsonOutput.toJson(DEFAULTS_JSON)) ])
 
+                            def copyArtifactSuccess = false
                             def downstreamJob = context.build job: downstreamJobName, propagate: false, parameters: buildJobParams
 
-                            if (downstreamJob.getResult() == 'SUCCESS') {
-                                // copy artifacts from build
-                                context.println '[NODE SHIFT] MOVING INTO CONTROLLER NODE...'
-                                context.node('worker') {
-                                    context.catchError {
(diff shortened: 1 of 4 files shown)


andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

pipelines/build/common/build_base_file.groovy
```diff
@@ -877,12 +879,19 @@ class Builder implements Serializable {
...
                                         }
+
+                                        copyArtifactSuccess = true
                                     }
+                            }
+                            context.println '[NODE SHIFT] OUT OF CONTROLLER NODE!'
+
+                            if ((downstreamJob.getResult() != 'SUCCESS' || !copyArtifactSuccess) && propagateFailures) {
+                                context.error("Propagating downstream job result: ${downstreamJobName}, Result: "+downstreamJob.getResult()+" CopyArtifactsSuccess: "+copyArtifactSuccess)
+                                if (copyArtifactSuccess && downstreamJob.getResult() == 'UNSTABLE' && (currentBuild.result == 'SUCCESS' || currentBuild.result == 'UNSTABLE' )) {
```
zdtsw, line 890: As for this logic: if pipeline has 3 downstream build jobs.
the 1st one finishes with **non-success** or **with success but failed to copy artifacts**, it immediately sets the pipeline status to **failure** or **unstable** , in this case, the whole pipeline will be exit i.e the other two downstream jobs get killed even they are still on-going?
Do i understand this correctly?

andrew-m-leonard: It doesn't exit, it just sets the pipeline result to Failure. The "jobs parallel" has to completely finish before the pipeline will exit.

andrew-m-leonard:
run tests

pipelines/build/common/build_base_file.groovy
andrew-m-leonard: So I confirmed, by aborting one, that the other builds continued, nothing was killed...

zdtsw (approved):

PR #642, opened 2023-03-07
Use (local) rhel7_build_image for Linux/s390x builds

sxa:
Needs https://github.com/adoptium/ci-jenkins-pipelines/pull/611 merged first
Fixes https://github.com/adoptium/infrastructure/issues/2731

PR #611, opened 2023-02-08
Do not issue docker pull if imagename contains rhel

sxa:
rhel images cannot be uploaded to dockerhub, so this is a test to see if we can use a local image on the machine. A final solution is likely to involve pushing it into a credentials-protected location, or having an additional parameter to skip the pull operation but that can be done as a subsequent step.

Signed-off-by: Stewart X Addison <sxa@redhat.com>

PR #652, opened 2023-03-09
Use rhel7_build_image for Linux/s390x builds (#642)

sxa:
Cherry picked across to release branch. May be subsequently superceded if we push the image to dockerhub (Ref https://github.com/adoptium/ci-jenkins-pipelines/pull/611#issuecomment-1458490322) but this puts it in line with master for now.


steelhead31 (approved):
LGTM

PR #667, opened 2023-03-21
Change build success to mean non failure

adamfarley:
Update the criteria for "failed build" to mean a build that literally has the failure status. This prevents downstream test timeouts and instability from affecting the failure rate of builds in the status reporting tooling, as the current criteria for failure is "anything but success". The job status of test jobs now affects the status of the build job that launched them, so this change helps clarify whether failures are associated with the build itself.

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -149,7 +149,7 @@ node('worker') {
                                     pipelineBuildJobsJson.each { buildJob ->
                                         if (buildJob.buildName.contains(buildVariant)) {
                                             buildJobNumber += 1
-                                            if (buildJob.buildResult.equals('FAILED')) {
+                                            if (buildJob.buildResult.equals('FAILURE')) {
                                                 buildJobFailure += 1
                       } else {
                                                 buildJobComplete += 1
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -90,7 +90,7 @@ node('worker') {
                         if (!foundNightly) {
                             def pipeline_id = null
                             def pipelineUrl
-                            def buildJobSuccess = 0
+                            def buildJobComplete = 0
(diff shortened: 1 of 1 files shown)


smlambert:
Another approach to this would be to change https://github.com/adoptium/ci-jenkins-pipelines/blob/59383b6ab5292447f37061ba1f51cd8ec6c6a5df/tools/nightly_build_and_test_stats.groovy#L145 to use a regex to find only the build jobs ( something like `${trssUrl}/api/getAllChildBuilds?parentId=${pipeline_id}\\&buildNameRegex=^jdk.*-temurin` )

much like is done for filtering for only Test jobs at https://github.com/adoptium/ci-jenkins-pipelines/blob/59383b6ab5292447f37061ba1f51cd8ec6c6a5df/tools/nightly_build_and_test_stats.groovy#L125.

This may be also how to add support for grouping and reporting on installer related jobs within the pipeline, etc.  Does enforce the need for naming schemas for builds / child jobs.

andrew-m-leonard:
@adamfarley Hmm, second thoughts, maybe above is too complicated. Lets keep it simple with what you have. Basically the Jenkins status of the build, that includes propagated test status, is essentially what we need, and if the status is low it indicates failing test jobs which we need to get to the bottom of, so is a good indicator we need to look at things.
We need to get those test job failures sorted.... to get the %'ages above 95%...

andrew-m-leonard:
> So how do we define/track success?

Sucess is ultimately the Build job is Green, as now all jenkins child jobs propagate their result

adamfarley:
@smlambert - I believe we already filter the jobs in a different way, as the only direct children of the pipeline job are the build jobs.

Would you be ok with reviewing this change as a short-term enhancement, while we discuss a bigger revamp [here](https://github.com/adoptium/ci-jenkins-pipelines/issues/674)?

smlambert (approved):

jobwe (commented):
tex reading

jobwe (commented):
nex bot

PR #681, opened 2023-04-11
Block aqa and tck tests if smoke tests fail or are unstable.

sophia-guo:
Close #121 

Signed-off-by: Sophia Guo <sophia.gwf@gmail.com>

PR #687, opened 2023-04-12
Copy downstream test jobs' Tap files no matter test jobs' result

### LINKED ISSUE #683: openjdk_build_pipeline jobs can't gather all downstream test job's tapfiles
openjdk_pipeline jobs can't gather all downstream test job's tapfiles. Take https://ci.adoptium.net/job/build-scripts/job/release-openjdk20-pipeline/5/ for example run verification https://ci.adoptium.net/view/Test_grinder/job/TAP_verification/24/console from console output some tap files are missing. Some jobs are not available now. From jobs available  tap files are missing if 

- test jobs result is aborted. https://ci.adoptium.net/job/build-scripts/job/jobs/job/release/job/jobs/job/jdk20/job/jdk20-release-windows-x64-temurin/5/ doesn't have extended.openjdk tapfile, the downstream job is aborted https://ci.adoptium.net/job/Test_openjdk20_hs_extended.openjdk_x86-64_windows/30/
- test jobs result is unstable. For example https://ci.adoptium.net/job/build-scripts/job/jobs/job/release/job/jobs/job/jdk20/job/jdk20-release-alpine-linux-x64-temurin/4/ doesn't have extended.openjdk tapfiles. THe downstream job is unstable  https://ci.adoptium.net/job/Test_openjdk20_hs_extended.openjdk_x86-64_alpine-linux/29/

sophia-guo:
Copy downstream test jobs' Tap files no matter test jobs' result

- Set propagate=false to access build job return object
- Explicitly set currentBuild.result by downstream job result
- AllowEmptyArchive so if no artifacts copied will not fail the build

Fix #683 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -441,50 +441,45 @@ class Build {
                                 }
                             }
                         }
-                        context.catchError {
-                            def testJob = context.build job: jobName,
-                                            propagate: true,
-                                            parameters: [
-                                                context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
-                                                context.string(name: 'UPSTREAM_JOB_NAME', value: "${env.JOB_NAME}"),
-                                                context.string(name: 'SDK_RESOURCE', value: 'upstream'),
-                                                context.string(name: 'JDK_REPO', value: jdkRepo),
-                                                context.string(name: 'JDK_BRANCH', value: jdkBranch),
-                                                context.string(name: 'OPENJ9_BRANCH', value: openj9Branch),
-                                                context.string(name: 'LABEL_ADDITION', value: additionalTestLabel),
-                                                context.booleanParam(name: 'KEEP_REPORTDIR', value: keep_test_reportdir),
-                                                context.string(name: 'PARALLEL', value: parallel),
-                                                context.string(name: 'NUM_MACHINES', value: "${numMachinesPerTest}"),
-                                                context.booleanParam(name: 'USE_TESTENV_PROPERTIES', value: useTestEnvProperties),
-                                                context.booleanParam(name: 'GENERATE_JOBS', value: aqaAutoGen),
-                                                context.string(name: 'ADOPTOPENJDK_BRANCH', value: aqaBranch),
-                                                context.string(name: 'ACTIVE_NODE_TIMEOUT', value: "${buildConfig.ACTIVE_NODE_TIMEOUT}"),
-                                                context.booleanParam(name: 'DYNAMIC_COMPILE', value: DYNAMIC_COMPILE)],
-                                            wait: true
-                            context.node('worker') {
-                                def result = testJob.getResult()
-                                context.echo " ${jobName} result is ${result}"
-                                if (result == 'SUCCESS' || result == 'UNSTABLE') {
-                                    context.sh 'rm -f workspace/target/AQAvitTaps/*.tap'
-                                    try {
-                                        context.timeout(time: 2, unit: 'HOURS') {
-                                            context.copyArtifacts(
-                                                projectName:jobName,
-                                                selector:context.specific("${testJob.getNumber()}"),
-                                                filter: "**/${jobName}*.tap",
-                                                target: 'workspace/target/AQAvitTaps/',
-                                                fingerprintArtifacts: true,
-                                                flatten: true
-                                            )
-                                        }
-                                    } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


karianna (approved):

smlambert (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -441,50 +441,45 @@ class Build {
...
+                                            context.booleanParam(name: 'KEEP_REPORTDIR', value: keep_test_reportdir),
+                                            context.string(name: 'PARALLEL', value: parallel),
+                                            context.string(name: 'NUM_MACHINES', value: "${numMachinesPerTest}"),
+                                            context.booleanParam(name: 'USE_TESTENV_PROPERTIES', value: useTestEnvProperties),
+                                            context.booleanParam(name: 'GENERATE_JOBS', value: aqaAutoGen),
+                                            context.string(name: 'ADOPTOPENJDK_BRANCH', value: aqaBranch),
+                                            context.string(name: 'ACTIVE_NODE_TIMEOUT', value: "${buildConfig.ACTIVE_NODE_TIMEOUT}"),
+                                            context.booleanParam(name: 'DYNAMIC_COMPILE', value: DYNAMIC_COMPILE)],
+                                        wait: true
+                        currentBuild.result = testJob.getResult()
```
andrew-m-leonard, line 464: This is not going to propagate correctly as it will report the last testJob result for the build job, hence a Success would overwrite a Failure, also we ought to report a failure on copy artifact failure?

I think we need something like (copied from similar code [here](https://github.com/adoptium/ci-jenkins-pipelines/blob/4d6f5e2f22e7c5330da9791ef930f3ddf65624e8/pipelines/build/common/build_base_file.groovy#L888)):
```
                            if ((testJob.getResult() != 'SUCCESS' || !copyArtifactSuccess) && propagateFailures) {
                                context.error("Propagating downstream job result: ${jobName}, Result: "+testJob.getResult()+" CopyArtifactsSuccess: "+copyArtifactSuccess)
                                if (copyArtifactSuccess && testJob.getResult() == 'UNSTABLE' && (currentBuild.result == 'SUCCESS' || currentBuild.result == 'UNSTABLE' )) {
                                    currentBuild.result = 'UNSTABLE'
                                } else {
                                    currentBuild.result = 'FAILURE'
                                }
                            }

```

andrew-m-leonard: Also, is this needed for this release?

andrew-m-leonard (changes requested):

pipelines/build/common/openjdk_build_pipeline.groovy
sophia-guo: The job status can only be set to worse than it https://github.com/jenkinsci/jenkins/blob/213363d387736874f1d14d83e57347f757f3ed4f/core/src/main/java/hudson/model/Run.java#L462-L466. That is **Success would overwrite a Failure**.  I also verified here https://ci.adoptium.net/view/work-in-progress/job/Sophia_pipeline/327/console

sophia-guo: 'copy artifact failure ' could happen if there is no artifacts to copy ( aborted or other exceptions), which is expected and has been reported by catch block
 ```catch (Exception e) {
                                context.echo "Cannot run copyArtifacts from job ${jobName}. Exception: ${e.message}. Skipping copyArtifacts..."
                            }```

sophia-guo: For aborted test jobs there might have artifacts ( parallel, one child job aborted, others good) or might not, which is reflected by test jobs status, so no need to report it specifically.

sophia-guo: Yes, I think both https://github.com/adoptium/ci-jenkins-pipelines/pull/681 and this PR are needed for release.

andrew-m-leonard: Interesting, never knew that was a hidden behaviour of assigning to currentBuild.result, that is illogical! but as you say does seem to be what happens. To me if I code a statement to assign result to be Success, it should be! If I were to code otherwise i'd code as result = WorseOf(current, Success) not hide the magic!!
Anyway, so yes lets go with that.

andrew-m-leonard (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
sophia-guo: Yes, I was going to do same thing  like WorseThan() and found it's unnecessary, which I think it is reasonable as it's the same logic when using propagate=true. With propagate=true the build status will only be set if downstream job status is worse.

PR #688, opened 2023-04-13
Allow eclipse-temurin-compliance-bot read access to release build job artifacts

andrew-m-leonard:
The ci.eclipse.org/temurin-compliance needs read access to build job artifacts for triggered TCK execution.


gdams (approved):

zdtsw (approved):

sxa (approved):
Approving on the basis that the reason this is now necessary is due to the permissions not being as restrictive as they should have been previously so this wasn't an issue in earlier releases.

PR #695, opened 2023-04-28
Change CI URL from adoptopenjdk -> adoptium

sxa:
Since the server was moved the curl command doesn't process the redirect so the download of gcc is failing.
This fixes it.

### DIFF
--- a/tools/reproduce_comparison/compareLinux.sh
+++ b/tools/reproduce_comparison/compareLinux.sh
@@ -49,7 +49,7 @@ if [ ! -r /usr/lib/jvm/jdk-${BOOTJDK_VERSION}/bin/javac ]; then
  echo Retrieving boot JDK $BOOTJDK_VERSION && mkdir -p /usr/lib/jvm && curl -L "https://api.adoptopenjdk.net/v3/binary/version/jdk-${BOOTJDK_VERSION}/linux/${NATIVE_API_ARCH}/jdk/hotspot/normal/adoptopenjdk?project=jdk" | (cd /usr/lib/jvm && tar xpzf -)
 fi
 if [ ! -r ${LOCALGCCDIR}/bin/g++-${GCCVERSION} ]; then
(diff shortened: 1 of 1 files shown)


tools/reproduce_comparison/compareLinux.sh
```diff
@@ -49,7 +49,7 @@ if [ ! -r /usr/lib/jvm/jdk-${BOOTJDK_VERSION}/bin/javac ]; then
  echo Retrieving boot JDK $BOOTJDK_VERSION && mkdir -p /usr/lib/jvm && curl -L "https://api.adoptopenjdk.net/v3/binary/version/jdk-${BOOTJDK_VERSION}/linux/${NATIVE_API_ARCH}/jdk/hotspot/normal/adoptopenjdk?project=jdk" | (cd /usr/lib/jvm && tar xpzf -)
 fi
 if [ ! -r ${LOCALGCCDIR}/bin/g++-${GCCVERSION} ]; then
-  echo Retrieving gcc $GCCVERSION && curl https://ci.adoptopenjdk.net/userContent/gcc/gcc$(echo $GCCVERSION | tr -d .).`uname -m`.tar.xz | (cd /usr/local && tar xJpf -) || exit 1
+  echo Retrieving gcc $GCCVERSION && curl -L https://ci.adoptopenjdk.net/userContent/gcc/gcc$(echo $GCCVERSION | tr -d .).`uname -m`.tar.xz | (cd /usr/local && tar xJpf -) || exit 1
```
karianna, line 52: should we not point directly to adoptium?

karianna (changes requested):

karianna (approved):

tools/reproduce_comparison/compareLinux.sh
gdams, line 52: As far as I can tell this PR doesn't do what is says? Anyway here's the change I'm guessing we need

```suggestion
  echo Retrieving gcc $GCCVERSION && curl -L https://ci.adoptium.net/userContent/gcc/gcc$(echo $GCCVERSION | tr -d .).`uname -m`.tar.xz | (cd /usr/local && tar xJpf -) || exit 1
```

sxa: The `-L` will allow it to redirect correctly, but yes it makes sense to have the name changed so it's trying to pull from the latest URL. Suggestion accepted.

gdams (approved):

Issue #706, opened 2023-05-30
evaluation-openjdk21-pipeline needs creating


andrew-m-leonard:
Now we are doing jdk-21 nightlies for the mainstream platforms, we should be doing the evaluation ones for jdk-21 as well.
https://ci.adoptium.net/job/build-scripts/

PR #707, opened 2023-05-30
Update evaluation_pipeline_generator.groovy so it generates "tip" version

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/706

Refactored slightly so it loops through versions, skipping retiredVersions, just like in https://github.com/adoptium/ci-jenkins-pipelines/blob/86519fe4d5b73269cb179333fbf7b0ab9ec71cd6/pipelines/build/regeneration/build_pipeline_generator.groovy#L51

### DIFF
--- a/pipelines/build/regeneration/evaluation_pipeline_generator.groovy
+++ b/pipelines/build/regeneration/evaluation_pipeline_generator.groovy
@@ -189,8 +189,12 @@ node('worker') {
                                 targetEvaluation = load nonUFile2
                             }
                         } catch (NoSuchFileException e2) {
-                            throw new Exception("[ERROR] enable to load jdk${javaVersion}u_evaluation.groovy nor jdk${javaVersion}_evaluation.groovy does not exist!")
+                            println "[WARNING] No evaluation config found for JDK${javaVersion} in the User's or Adopt's repository. Skipping generation..."
+                            // break and move to next element in the loop
+                            // groovylint-disable-next-line
+                            return
                         }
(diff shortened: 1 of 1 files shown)


pipelines/build/regeneration/evaluation_pipeline_generator.groovy
```diff
@@ -48,7 +48,7 @@ node('worker') {
         }
 
         timestamps {
-            def validVersion = [8, 11, 17, 20]
+            def retiredVersions = [9, 10, 12, 13, 14, 15, 16, 18, 19]
```
karianna, line 51: Hmm, do we need a shared properties ofr config file to track this sort of thing across our logic?

karianna (approved):

pipelines/build/regeneration/evaluation_pipeline_generator.groovy
andrew-m-leonard: Perhaps yes. Not quite sure what a "shared properties ofr config file" is ? @karianna 
Using the jenkins-helper might be possible, possibly add to : https://github.com/adoptium/jenkins-helper/blob/master/vars/Constants.groovy

zdtsw: think we also need to update documentation (related to post release work) when we retire a release.

zdtsw (approved):

PR #712, opened 2023-06-02
Fix user ci pipeline setup usage

andrew-m-leonard:
Noticed a problem in build-pipeline-generator when trying to use a User config repository, the fallback to use the weekly job template has the wrong json value.

It should be "weekly", not "weeklyTemplatePath": 
https://github.com/adoptium/ci-jenkins-pipelines/blob/b743ba02353aeef2942dae87c20173c19b7f79ee/pipelines/defaults.json#L17


zdtsw (approved):
/lgtm

zdtsw:
> Hi Andrew. Your change LGTM. Should evaluation_pipeline_generator.groovy also be modified in the same way?

yep

andrew-m-leonard:
> Hi Andrew. Your change LGTM. Should evaluation_pipeline_generator.groovy also be modified in the same way?

@adamfarley yes, good spot. Thanks

adamfarley (approved):
LGTM :)

Issue #717, opened 2023-06-08
Upstream jdk-21 fork, and head moving to jdk-22


andrew-m-leonard:
Jdk-21 is forking to jdk21 repo June 8th 2023: https://mail.openjdk.org/pipermail/jdk-dev/2023-June/007894.html
Jdk(head) will become jdk-22

Specific jdk21 build jobs will need creating:
Tasks: https://github.com/adoptium/temurin-build/wiki/Setting-up-new-build-jobs,-mirrors,...-when-a-new-openjdk-version-is-branched-from-jdk(head)

andrew-m-leonard:
Raised: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3252

gdams:
@andrew-m-leonard, reminder that next time you need a new repo you can raise a PR to https://github.com/adoptium/.eclipsefdn/blob/main/otterdog/adoptium.jsonnet

andrew-m-leonard:
> @andrew-m-leonard, reminder that next time you need a new repo you can raise a PR to https://github.com/adoptium/.eclipsefdn/blob/main/otterdog/adoptium.jsonnet

ah thanks @gdams , not sure how that works? Once that file is updated, does it automatically create the repo?

and do I need to retrospectively add adoptium/jdk21 to it?

gdams:
> > @andrew-m-leonard, reminder that next time you need a new repo you can raise a PR to https://github.com/adoptium/.eclipsefdn/blob/main/otterdog/adoptium.jsonnet
> 
> ah thanks @gdams , not sure how that works? Once that file is updated, does it automatically create the repo?
> 
> and do I need to retrospectively add adoptium/jdk21 to it?

It will eventually be automatic but right now the EF have to run a script once they've merged the changes. And no, you don't need to add jdk21, it will be automatically created in the jsonnet next time it's run.

PR #719, opened 2023-06-09
For JDK head, use alternate file/release name

sxa:
Part of providing a solution to https://github.com/adoptium/temurin-build/issues/3355

This changes the format of the "nightlies" from: `yyyy-HH-dd-HH-mm` to `ea.21+23` which will be reflected in the filename and also in the published tag. Note that `ea.21+23` was chosen instead of `21+23.ea` as this fits into the [existing regex](https://github.com/adoptium/github-release-scripts/blob/ba2f0167d794c1cf29314292265045453a8ce7bc/sbin/Release.sh#L44).

This is currently in draft for multiple reasons:
1. I have not tested the end to end code
2. We are in the process of [splitting out JDK21 from JDK](https://github.com/adoptium/ci-jenkins-pipelines/issues/717), so headVersion is probably not what we want now
3. Looking for validation on the formatting of the filenames and releases
4. Because of the criteria I've used in this PR, perhaps we should be adjusting the schedule to disable nightlies from new tag before implementing this? Or have a different criteria for these (e.g. "not release but tag specified"). For now we could manually trigger, but a variant of https://github.com/adoptium/mirror-scripts/blob/master/triggerReleasePipeline.sh (called by e.g. https://ci.adoptium.net/job/build-scripts/job/utils/job/releaseTrigger_jdk20u/) which just builds any level that hasn't been built previously rather than working on a static level would be ideal.

_NOTE: If we had such a trigger another option would be to still run the nightly build/test cycles as-is but run them as "nightly without publish" so as not to pollute the binaries repositories._

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -664,7 +664,6 @@ class Builder implements Serializable {
         return "jdk${number}"
     }
 
-
     /*
     Returns the downstream build job's type by checking job folder's path
     can be "evaluation" or "release" or null (in this case it is for the nightly or pr-tester)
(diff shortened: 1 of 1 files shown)


sxa:
run tests

sxa:
run tests

PR #720, opened 2023-06-09
Add jdk22 pipeline config

andrew-m-leonard:
jdk(head) is now jdk-22, add the config.

For the moment disable scheduling until we have determined a nightly schedule for it.


gdams (approved):

sxa (approved):
LGTM, and noting that with the changes in https://github.com/adoptium/ci-jenkins-pipelines/pull/719 and as per the PMC call this week we could have a trigger that runs when new `_adopt` tags are created in the repository instead of having it regularly scheduled.

Issue #726, opened 2023-06-13
IndividualBuildConfigTest does not test config members correct


andrew-m-leonard:
Test: https://github.com/adoptium/ci-jenkins-pipelines/blob/04dda51e33ada01bca22e75626d2d4c7f1eb4ebb/pipelines/src/test/groovy/IndividualBuildConfigTest.groovy#L61
Does not actually verify the values in the config Map are valid in IndividualBuildConfig, as it uses toJson() on the IndividualBuildConfig object, which will only have what is valid!

PR #727, opened 2023-06-13
Assert IndividualBuildConfig values are defined in the jenkins-helper

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/726

Correct IndividualBuildConfig map assertion to detect missing/invalid members, not defined in https://github.com/adoptium/jenkins-helper/blob/783ece1c4666bc4092494438e62775cc6701025d/src/common/IndividualBuildConfig.groovy#L19


steelhead31 (approved):
LGTM.

PR #520, opened 2022-11-29
feat: add support for evaluation jobs/pipelines

zdtsw:
**Highlight change:**

split current jdkX.groovy config into two parts:
- jdkX.groovy will be used for nightly builds
- jdkX_evaluation.groovy will also be run as the same time as jdkX.groovy, but only contain unstable/non-temurin variant builds

daily build stats should only count whats in the jdkX.groovy, (can be another type of build stats if required for evaluation, and we already have ones for openj9 etc variants)
weekly archive should only use jdkX.groovy as input and archive results
introduce new weekly-evaluation use jdkX_prototpye.groovy as input for weekly cycle long test.

**Detail Config updates**
aarch64 windows is added back as part of evaluation
openj9, etc are moved into evaluation
official release of temurin should refer to whats stated in jdkX_release.groovy 

in short:

- jdkX_release.groovy is the one we used to include what we release to customer, it is only used for release time
- jdkX.groovy included everything from jdkX_release + some others we do not release but should be counted into stats
- jdkX_evaluation.groovy has config not in either of above two config files: not ready to be included into nightly, not temurin builds, etc 

**Trigger updates**
- `openjdkX-pipeline`  relesaseType to `Nightly` use `triggerSchedule_nightly`
- `weekly-openjdkX-pipeline` relesaeType to "`Release`"  t`riggerSchedule_weekly`

-   `weekly-prototpye-openjdkX-pipeline` releasetype to `Nightly Not Publish` use `triggerSchedule_weekly_evaluation`: this will disable the remotetrigger to JCK (isRelease = false)
-  `evaluation-openjdkX-pipeline`  releaseType to `Nightly Not Publish` use t`riggerSchedule_evaluation`

first two share the same config file and  targetConfiguration, later two share the same config file and targetConfiguration

**Test results**
Test pipeline generator: https://ci.adoptopenjdk.net/view/work-in-progress/job/wen-build-pipeline-generator/
- generates build-scripts-testing/evaluation-openjdk20-pipeline

Tet job generator: https://ci.adoptopenjdk.net/view/work-in-progress/job/wen-evaluation-pipeline_jobs_generator_jdk/
- generates build-scripts-testing/evaluation/jobs/jdk20/jdk20-evaluation-linux-aarch64-hotspot etc

--------------------------------------------------------------------------------------------------------------------------------------------------
result job to create downstream evaluation jobs
[...]

PR #729, opened 2023-06-14
Removing chance jobReleaseType can cause duplicate values

adamfarley:
Sometimes the value of jobReleaseType is "Nightly". This creates confusion, and ambiguity, as people running the pipelines don't know which "Nightly" to use when kicking off a pipeline.

This change removes any duplicate strings.

Adding Wen as the original [implementor](https://github.com/adoptium/ci-jenkins-pipelines/pull/520) of the code I'm tweaking, and Stewart as the person who noticed the issue.


pipelines/jobs/pipeline_job_template.groovy
```diff
@@ -94,7 +94,7 @@ pipelineJob("${BUILD_FOLDER}/${JOB_NAME}") {
...
         stringParam('buildConfigFilePath', '', "Relative path to where the jdkxx_pipeline_config.groovy file is located. It contains the build configurations for each platform, architecture and variant.<br>Default: <code>${defaultsJson['configDirectories']['build']}/jdkxx_pipeline_config.groovy</code>")
-        choiceParam('releaseType', [jobReleaseType, 'Nightly', 'Nightly Without Publish', 'Weekly', 'Release'], 'Nightly - release a standard nightly build.<br/>Nightly Without Publish - run a nightly but do not publish.<br/>Weekly - release a standard weekly build, run with extended tests.<br/>Release - this is a release, this will need to be manually promoted.')
+        choiceParam('releaseType', [jobReleaseType, 'Nightly', 'Nightly Without Publish', 'Weekly', 'Release'].unique(), 'Nightly - release a standard nightly build.<br/>Nightly Without Publish - run a nightly but do not publish.<br/>Weekly - release a standard weekly build, run with extended tests.<br/>Release - this is a release, this will need to be manually promoted.')
```
zdtsw, line 97: does it even support function to be called on this level ?
or if  you have a screenshot after your change

zdtsw: i would say to call unique() is better done in https://github.com/adoptium/ci-jenkins-pipelines/blob/22a9cf2a46d787bcf5fa27d5f88317974f5fdb9f/pipelines/jobs/pipeline_job_template.groovy#L13 after convert jobReleaseType to list jobReleaseTypeList by adding the rest into it

adamfarley: Heya Wen,

I think you can use unique this way because you can see [here](https://riptutorial.com/groovy/example/18010/remove-duplicates) that unique works on a list of strings, and you can see [here](https://blog.mrhaki.com/2010/01/groovy-goodness-removing-duplicate.html) (line 17) that Groovy allows you to call methods directly on an array without assigning the array to a named variable.

I did try something similar to what you suggest, but this seemed simpler. Plus, it keeps all the data in one place, next to the choice parameter we're populating.

andrew-m-leonard:
@adamfarley The key thing is the "default" value selected for the choice is the jobReleaseType value, hence why it was duplicated as the 1st item, which is what Jenkins chooses as the "default"

andrew-m-leonard (commented):
Assuming unique() works there, should be fine.
You need to re-base your branch I think for "Test" failure

adamfarley:
> @adamfarley The key thing is the "default" value selected for the choice is the jobReleaseType value, hence why it was duplicated as the 1st item, which is what Jenkins chooses as the "default"

That makes sense. I thought it was to allow a dynamic value in the list that may not be predicted in advance (like "Nightly" and "Release").

Will create a smarter solution to address Andrew and Wen's concerns. Adding WIP to the title.

andrew-m-leonard (approved):
lgtm

sxa (approved):
SGTM

PR #731, opened 2023-06-15
Disable reproducible comparision for pr-tester and release per platform

sophia-guo:
Setting ENABLE_REPRODUCIBLE_COMPARE as boolean
Other updates: JDKs path, dir, label

Related https://github.com/adoptium/temurin-build/issues/3245

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -125,7 +125,7 @@ class Builder implements Serializable {
         if (additionalBuildArgs) {
             buildArgs += ' ' + additionalBuildArgs
         }
-        def enableReproducibleCompare = getReproducibleCompare(platformConfig, variant)
+        def enableReproducibleCompare = isEnableReproducibleCompare(platformConfig, variant)
         def testList = getTestList(platformConfig, variant)
 
         def dynamicTestsParameters = getDynamicParams(platformConfig, variant)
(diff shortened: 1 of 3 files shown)


karianna (approved):

pipelines/build/common/build_base_file.groovy
```diff
@@ -223,15 +223,19 @@ class Builder implements Serializable {
...
-            if (isMap(configuration.reproducibleCompare)) {
-                reproducibleCompare = (configuration.reproducibleCompare as Map).get(variant)
-            } else {
-                reproducibleCompare = configuration.reproducibleCompare
-            }
-            if (reproducibleCompare != null) {
-                enableReproducibleCompare = reproducibleCompare
+        if ( env.JOB_NAME.contains('pr-tester') || env.JOB_NAME.contains('release')) {
+            enableReproducibleCompare = false
+        } else {
```
zdtsw, line 228: personally speaking, I feel this logic should be added when the job is generated not the build is called. e.g in [getReproducibleCompare](https://github.com/adoptium/ci-jenkins-pipelines/blob/b7e00a70ab650a485c47120deab22cbc001e9ad4/pipelines/build/common/config_regeneration.groovy#L342)

pipelines/build/common/build_base_file.groovy
```diff
@@ -223,15 +223,19 @@ class Builder implements Serializable {
     */
     Boolean getReproducibleCompare(Map<String, ?> configuration, String variant) {
```
zdtsw, line 224: another thing is about the function name. 
maybe something like `isEnableReproducibleCompare` is better than `getReproducibleCompare` ?

tools/reproduce_comparison/Jenkinsfile
```diff
@@ -4,7 +4,7 @@ Jenkins job does reproducible build compare.
 import groovy.json.JsonOutput
 
 pipeline {
-    agent any
+    agent { label 'dockerBuild&&linux&&x64' }
```
zdtsw, line 7: fine with this, but if mac and windows logic are enabled later, we need find a better label combination or even move agent setting into later stage .

tools/reproduce_comparison/Jenkinsfile
```diff
@@ -59,7 +59,7 @@ pipeline {
                     jsonJobParams.BUILD_CONFIGURATION.SCM_REF = scmRef
                     jsonJobParams.BUILD_CONFIGURATION.BUILD_REF = buildRef
                     jsonJobParams.BUILD_CONFIGURATION.BUILD_ARGS += " --build-reproducible-date ${buildTimeStamp}"
-                    jsonJobParams.BUILD_CONFIGURATION.ENABLE_REPRODUCIBLE_COMPARE = "false"
+                    jsonJobParams.BUILD_CONFIGURATION.ENABLE_REPRODUCIBLE_COMPARE = false
```
zdtsw, line 62: think the reason with "" is to treat it as String, since textParam only support String.  Can test and see if it work here

andrew-m-leonard (approved):

tools/reproduce_comparison/Jenkinsfile
sophia-guo: ENABLE_REPRODUCIBLE_COMPARE is defined as Boolean https://github.com/adoptium/jenkins-helper/blob/master/src/common/IndividualBuildConfig.groovy#L55

pipelines/build/common/build_base_file.groovy
sophia-guo: Configurations in https://github.com/adoptium/ci-jenkins-pipelines/blob/b7e00a70ab650a485c47120deab22cbc001e9ad4/pipelines/build/common/config_regeneration.groovy are used as a placeholder it could be overwritten during the build time.

Issue #773, opened 2023-08-11
buildpipeline generator set trigger even if triggerSchedule_nightly not in the config file


andrew-m-leonard:
jdk-21 pipeline job trigger was randomly getting set to a triggerSchedule_nightly, even though the jdk21.groovy config has no value....?

It seems in Jenkins groovy, a local "def target" variable that is not initialized within an iterative loop, will still have the values from the previous iteration! Hence, the jdk21 iteration was picking up the jdk17u or jdk20u schedule!!

PR #774, opened 2023-08-11
Ensure target is reset before loading a jdk pipeline config file

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/773

When pipeline config is loaded via a groovy "target = load <config>", target is set to the current groovy binding, and all the variables loaded are set in the groovy variable binding.
When a 2nd load is performed the target return is the same "current groovy binding", thus any previous version loaded variable values will still be set... hence why the jdk21 schedule was being set to the jdk17u schedule!

This PR ensures at the end of each iteration the target binding values are reset.

### DIFF
--- a/pipelines/build/regeneration/build_pipeline_generator.groovy
+++ b/pipelines/build/regeneration/build_pipeline_generator.groovy
@@ -166,7 +166,8 @@ node('worker') {
                     releaseType         : 'Nightly' 
                 ]
 
-                def target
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

tellison (approved):

sxa (approved):
I'd question whether a try/catch is the nicest way to achieve this, but if we're confident that it fixes the problem I'll give my tentative approval on it (But to be clear I'm not an expert in this area of the code)

Issue #789, opened 2023-08-22
jdk21 and 17 evaluation schedules not being set correctly


andrew-m-leonard:
The evaluation pipeline generation needs this fix as well: https://github.com/adoptium/ci-jenkins-pipelines/commit/cedc5010be4424a0da5f85c7c58f4ca404841539
to ensure loaded config doesn't pickup old values...

PR #790, opened 2023-08-22
Ensure evaluation pipeline config load resets groovy binding before the load

andrew-m-leonard:
Ensure evaluation pipeline config load resets groovy binding before the load, so any old values are not picked up.
Same fix as https://github.com/adoptium/ci-jenkins-pipelines/commit/cedc5010be4424a0da5f85c7c58f4ca404841539
Also fixed typo in jdk17u schedule.

Fixes: https://github.com/adoptium/ci-jenkins-pipelines/issues/789


karianna (approved):

sxa (approved):
I don't know enough about the way these pipelines work to give a full code review based on the description, but this superficially sounds reasonable and if you believe it fixes the pipeline triggers then it shouldn't make things any worse, so I'm approving :-)

Issue #798, opened 2023-08-30
Limited visibility of published nightly/EA triggered builds missing artifacts


andrew-m-leonard:
If a nightly build fails to publish a given artifact, or maybe a platform failed to build, then for the regular scheduled nightly, it would just be re-built next time.. However, with the new upstream build tag triggered builds, there is no next schedule, so in that scenario, we need to be aware that a platform failed, or an artifact failed to publish?

Issue #787, opened 2023-08-21
Daily Slack "stale build" status needs to handle tag-triggered builds


andrew-m-leonard:
jdk-21/22 beta builds are triggered by upstream "tags", and so may not build for a while if no upstream builds occur. The daily Slack status checks for a fixed "5 days", this needs changing to check if the latest build tag has been built.

Additionally, we should verify the release assets all exist that should do, in case of a failed platform or a failed upload of an asset file

PR #799, opened 2023-08-30
Refactor daily status health check to handle triggered builds and verify published assets

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/787
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/798

Update the daily Slack build pipeline health status:
- Handle build tag triggered builds that do not have a max stale days.
- For build tag triggered builds, confirm the latest published binary is the latest upstream openjdk build tag.
- For all versions, verify the latest published (according to API) binaries have all the required artifacts.
- Simplify the build&test success rating to simply find the last pipeline build from the trss API
- Parameterise the featureReleases and optional "tipRelease" versions

Added job Parameters:
FEATURE_RELEASES: List of jdkNNu releases to get status of
TIP_RELEASE: Optional jdk(head) release version (eg.jdk22)
BINARIES_REPO: github.com binaries repo template (eg.adoptium/temurin_NN_-binaries)
BUILD_CONFIG_URL: URL of raw ci-jenkins-piplines pipeline configuration for obtaining required Arch's for version

Test Slack status example: https://adoptium.slack.com/archives/C598A7EMQ/p1693580500301949

Missing Windows SBOM artifacts discovered with this work, raised as issue: https://github.com/adoptium/temurin-build/issues/3464

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -454,10 +454,12 @@ echo 'Adoptium Latest Builds Success : *' + variant + '* => *' + overallNightlyS
                     def slackColor = 'good'
                     def health = "healthy"
                     def errorMsg = ""
-                    if (!latestTagBuilt) {
+                    def missingAssets = []
+                    if (status['assets'] != "Complete") {
                         slackColor = 'danger'
                         health = "unhealthy"
-                        errorMsg = "Expected build tag: "+status['expectedReleaseName']
+                        errorMsg = " Artifact status: "+status['assets']
+                        missingAssets = status['missingAssets']
                     } else if (maxDays <= days) {
                         slackColor = 'warning'
                         health = "unhealthy"
@@ -465,6 +467,9 @@ echo 'Adoptium Latest Builds Success : *' + variant + '* => *' + overallNightlyS
                     }
                     def fullMessage = "JDK ${featureRelease} latest pipeline publish status: ${health}. Build: ${releaseName}. Published: ${msg}.${errorMsg}"
                     echo "===> ${fullMessage}"
+                    if (missingAssets.size() > 0) {
+                        echo "==>     Missing artifacts: "+missingAssets
+                    }
                     // One slack message per JDK version:
                     //slackSend(channel: slackChannel, color: slackColor, message: fullMessage)
                 } else {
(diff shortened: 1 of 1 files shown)


jerboaa (commented):
Personally I find the added condition to mark the build status as unhealthy when a build artefact is missing a bit noisy. I'd be primarily interested in a basic "is there a new EA binary to test available" message, which is now hard to discern on this extended status. Perhaps that's a non-issue, and this new status will mostly work fine, but if it turns out to be red 99% of the time, it starts to become less informative and becomes noise.

Perhaps we want two status?

1. Is **a** binary available for testing (used to be Linux x86_64).
2. Are there any missing artefacts? If so, which.

Thoughts?

andrew-m-leonard:
> Personally I find the added condition to mark the build status as unhealthy when a build artefact is missing a bit noisy. I'd be primarily interested in a basic "is there a new EA binary to test available" message, which is now hard to discern on this extended status. Perhaps that's a non-issue, and this new status will mostly work fine, but if it turns out to be red 99% of the time, it starts to become less informative and becomes noise.
> 
> Perhaps we want two status answering the following to basic questions?
> 
>     1. Is **a** binary available for testing (used to be Linux x86_64)?
> 
>     2. Are there any missing artefacts? If so, which.
> 
> 
> Thoughts?

@jerboaa Thanks for the Feedback.
So this change is partly driven by are gradual change to only doing EA builds when the upstream tag gets published, and the importance of making people/us aware if there is something wrong/missing, with the latest available EA tag build, as no repeat build will occur.
The intention is this daily status should ALWAYS be "Healthy", and if not we act upon it quickly to make it so... Being Healthy, means ALL platforms are available at that latest build tag.
Previously, the "health" was just based on a publish x64Linux being available and published within last 5days, it gave no information on whether other platforms were also available.
So, the intention should be:
1. A "Healthy" indication that a binary for your chosen platform is available, and the name and link of that EA build
2. In the case of "Unhealthy" a list of the artifacts/platforms that are missing from that build. Which we will then act upon to resolve...

jerboaa:
> The intention is this daily status should ALWAYS be "Healthy"

Alright. If the intention is to keep this healthy, I have no concerns with this. We'll see how it goes.

karianna (approved):

PR #801, opened 2023-08-31
Reformat tags for hotspot builds to ensure they match upstream tags

adamfarley:
*_adopt tags don't match anything upstream, so I'm removing the "_adopt" bit when we're running a hotspot build.

Testing here: https://ci.adoptium.net/job/build-scripts/job/openjdk21-pipeline/177/

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -146,6 +146,12 @@ class Builder implements Serializable {
             cleanWsAfter = platformCleanWorkspaceAfterBuild
         }
 
+        // We need to ensure that _adopt is stripped from any tags used in hotspot variant builds, as *_adopt tags do not exist upstream.
+        def adjustedScmReference = scmReference
+        if (variant.equals("hotspot")) {
+            adjustedScmReference = scmReference - ('_adopt')
+        }
+
         return new IndividualBuildConfig(
             JAVA_TO_BUILD: javaToBuild,
             ARCHITECTURE: platformConfig.arch as String,
@@ -154,7 +160,7 @@ class Builder implements Serializable {
             TEST_LIST: testList,
             DYNAMIC_LIST: dynamicList,
             NUM_MACHINES: numMachines,
-            SCM_REF: scmReference,
+            SCM_REF: adjustedScmReference,
             BUILD_REF: buildReference,
             CI_REF: ciReference,
             HELPER_REF: helperReference,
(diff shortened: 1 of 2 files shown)


adamfarley:
Ok, typo caused a failure. Fixed the typo, and then decided that we don't want to put this logic within the build itself. Tidier if we put it in the pipeline job.

Change made. Running test now: https://ci.adoptium.net/job/build-scripts/job/openjdk21-pipeline/178/

andrew-m-leonard:
run tests

karianna:
Have pings the PMC on this one in case there's historical context

gdams:
This won’t work, the whole reason for the _adopt tag is that it contains any Adoptium patches that have been merged from the dev branch into the release branch.

andrew-m-leonard:
> This won’t work, the whole reason for the _adopt tag is that it contains any Adoptium patches that have been merged from the dev branch into the release branch.

@gdams for some context, this is "hotspot" builds, not "temurin" builds, see:
https://github.com/adoptium/ci-jenkins-pipelines/blob/8c0442b810fd456ff4027bc387c3059ce295ae9f/pipelines/jobs/configurations/jdk21.groovy#L24
We do a sanity build of upstream "hotspot" for jdk-21, which fails as obviously upstream openjdk don't have an _adopt tag..
If we think we should just not build this, then that's an option too? @sxa

gdams:
Ahh okay sorry that makes sense now.

gdams (approved):

sxa:
What is the problem that's being fixed here? That's not clear from the description. Is there something specific that's failing?

adamfarley:
Heya Stewart. The problem we're fixing here is that the tag-specific ea builds of jdk21 ([example](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21/job/jdk21-linux-aarch64-hotspot/61/)) are failing when building the hostpot (NOT Temurin) variant, because we're seeking a specific tagged level which IS present at Adoptium, but is NOT present when we clone directly from the upstream OpenJDK source repos.

In short, we need to remove the "_adopt" bit from hotspot-variant builds before they will work.

This change does that in a way that lets us continue to execute the hotspot builds with the same pipeline job as the Temurin builds.

adamfarley:
Looks like we have no remaining objections. Will squash and merge.

sxa:
> Heya Stewart. The problem we're fixing here is that the tag-specific ea builds of jdk21 ([example](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21/job/jdk21-linux-aarch64-hotspot/61/)) are failing when building the hotspot (NOT Temurin) variant

Makes sense - thanks for clarifying.

PR #802, opened 2023-09-01
Add weekly tests to ea pipelines

adamfarley:
To achieve the title goal, we should add a "Weekly without Publish" to the build pipelines.

In the long run we should find a neater way to run ea specific-tagged-level build testing. Perhaps when all of Adoptium's builds are ea (and nightly/weekly becomes an unused concept), we could have a single test list for all builds.

This seems a good solution in the short term though.

Testing here: https://ci.adoptium.net/job/build-scripts/job/openjdk21-pipeline/183/

And the "without publish" should work to prevent publishing by avoiding the triggering of this code [here](https://github.com/adoptium/ci-jenkins-pipelines/blob/36de9288e3eda05ba64498b824e9d4cb6a560137/pipelines/build/common/build_base_file.groovy#L1004).


karianna (approved):

andrew-m-leonard (approved):
lgtm

Issue #791, opened 2023-08-22
jdk17u & jdk21 comparable build job returns FAILURE status on difference found


andrew-m-leonard:
To be consistent with "test" type jobs, as long as the job completed, but maybe found comparable differences, then it should return UNSTABLE, rather than FAILURE

PR #807, opened 2023-09-06
Update reproduce_comparison to return UNSTABLE for differences found

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/791

Updated results to UNSTABLE if repro_compare returns a difference.
Also changed failures to un-tar to be FAILUREs.

Test compare: https://ci.adoptium.net/view/ReproducibleBuild/job/jdk21-mac-aarch64-temurin_reproduce_compare/7/

### DIFF
--- a/tools/reproduce_comparison/Jenkinsfile
+++ b/tools/reproduce_comparison/Jenkinsfile
@@ -161,7 +161,7 @@ pipeline {
                             }
                             if (rc != 0) {
                                 currentBuild.result = 'UNSTABLE'
-                                echo 'Failed: two builds are not the same! Please see the archived repro_diff.out.'
+                                echo 'Differences found: two builds are not the same! Please see the archived repro_diff.out.'
                             } else {
                                 echo 'Success: two builds are the same!'
                             }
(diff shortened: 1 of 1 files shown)


sophia-guo (commented):
it would be better if the return code of https://github.com/adoptium/temurin-build/blob/master/tooling/repro_compare.sh can be differentiated as 0 ( same ), 1( errors), 2 (different). Correspondingly the jenkins jobs status would be SUCCESS, FAILURE, UNSTABLE.

sophia-guo (approved):

andrew-m-leonard:
> it would be better if the return code of https://github.com/adoptium/temurin-build/blob/master/tooling/repro_compare.sh can be differentiated as 0 ( same ), 1( errors), 2 (different). Correspondingly the jenkins jobs status would be SUCCESS, FAILURE, UNSTABLE. If @andrew-m-leonard agree I can open a separate issue in temurin repo.

Yes please @sophia-guo good plan

smlambert:
This also somewhat relates to converting this to a real test case (where there are specific / defined error codes and messages).

karianna (approved):

PR #820, opened 2023-09-26
Rename jdk21 pipeline config to jdk21u

andrew-m-leonard:
Rename jdk21 pipeline job configuration to jdk21u.


sxa:
Hmm do really want to do a rename when we haven't actually shipped jdk21GA so there's always the possibility of a reason (albeit small)?

Regardless, we should ensure that nothing gets schedules to run and publish from jdk21u automatically at the moment as the latest nightly/EA really needs to stay pointing at jdk-21+35 for now until we so the release build.

sxa:
Also as a point of note; Please do not remove the JDK21 experimental build pipelines as I have some custom books in there that I don't want to lose.

steelhead31 (approved):
LGTM

sxa:
>  We have the v2023.09.01 branch with the release (jdk21) configuration:

Yep my primary concern here was to make sure that we have new jdk21u things generated and retain the existing jdk21 ones (so that branch could still use the old jobs if needed) and that we didn't rename/delete the existing ones

karianna (approved):

PR #825, opened 2023-10-10
Update x64 Mac build jobs to use macos11 label

Haroon-Khel:
In light of https://github.com/adoptium/infrastructure/issues/2536#issuecomment-1697270848 and https://github.com/adoptium/temurin-build/issues/3354, and now that https://github.com/adoptium/temurin-build/pull/3492 is in we can force the builds to run on Macos11 machines only. 

I am positive I have updated the correct files, but feel free to correct me if I have not.

There is one problem. This line, https://github.com/adoptium/ci-jenkins-pipelines/blob/5fcae13f466c8063ca572eaa8365c13548fede46/pipelines/build/common/config_regeneration.groovy#L465, will insist that the `NODE_LABEL` contains the architecture of the binary to build, in this case `x64`. So the final label will be `"NODE_LABEL": "macos11&&build&&mac&&x64"` which will prevent it from running on our arm mac build machines. 

An easy solution is to add `x64` labels to our build arm mac machines. A harder solution is to modify https://github.com/adoptium/ci-jenkins-pipelines/blob/5fcae13f466c8063ca572eaa8365c13548fede46/pipelines/build/common/config_regeneration.groovy#L465 to something more versatile which will not break the other build jobs, but at the moment I do not know this repo well enough for this to be a quick fix before the next release. 

@sxa @andrew-m-leonard thoughts?

>An easy solution is to add `x64` labels to our build arm mac machines.

I'm in favour of this

### DIFF
--- a/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
@@ -4,7 +4,7 @@ class Config11 {
         x64Mac    : [
                 os                  : 'mac',
                 arch                : 'x64',
-                additionalNodeLabels : 'macos10.14',
+                additionalNodeLabels : 'macos11',
                 test                : 'default',
                 configureArgs       : [
                         'openj9'      : '--enable-dtrace=auto --with-cmake',
--- a/pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk17u_pipeline_config.groovy
@@ -4,7 +4,7 @@ class Config17 {
         x64Mac    : [
                 os                  : 'mac',
                 arch                : 'x64',
-                additionalNodeLabels: 'macos10.14',
+                additionalNodeLabels: 'macos11',
                 additionalTestLabels: [
                         openj9      : '!sw.os.osx.10_11'
                 ],
--- a/pipelines/jobs/configurations/jdk20u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk20u_pipeline_config.groovy
@@ -4,7 +4,7 @@ class Config20 {
         x64Mac    : [
                 os                  : 'mac',
                 arch                : 'x64',
-                additionalNodeLabels: 'macos10.14',
+                additionalNodeLabels: 'macos',
                 additionalTestLabels: [
                         openj9      : '!sw.os.osx.10_11'
                 ],
--- a/pipelines/jobs/configurations/jdk20u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk20u_pipeline_config.groovy
@@ -4,7 +4,7 @@ class Config20 {
         x64Mac    : [
                 os                  : 'mac',
                 arch                : 'x64',
-                additionalNodeLabels: 'macos',
+                additionalNodeLabels: 'macos11',
                 additionalTestLabels: [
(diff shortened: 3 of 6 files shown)


sxa (changes requested):
I don't think this will work - it'll look for [macos11&&build&&mac&&x64](https://ci.adoptium.net/label/macos11&&build&&mac&&x64/) (which we don't have any of) instead of selecting an aarch64 build machine

sxa:
Take a look at the `crossCompile` sections in jdk11u for 32-bit arm and riscv64 for an example of how to make it run on a different machine. Hopefully something that that will work here too.

Haroon-Khel:
>Take a look at the crossCompile sections in jdk11u for 32-bit arm and riscv64 for an example of how to make it run on a different machine. Hopefully something that that will work here too.

I have added `x64` labels to the build-arm-mac machines so they now show up in https://ci.adoptium.net/label/macos11&&build&&mac&&x64/. Let's merge this pr and if we see stable builds then I can implement a more versatile solution

sxa (approved):
While it's not ideal having that label on a machine that is of a different architecture I'll approve this on the basis that this is an interim solution for next week's release and we will look at doing it properly for the next release. Let's also make sure that when we merge this we take the build label OFF the x64 ones (and log here which machines that's been done on) to ensure that we don't end up with a random selection of x64 or aarch64 systems used for builds which could affect reproducibility

andrew-m-leonard:
I think we might be able to use the crossCompile option, eg: https://github.com/adoptium/ci-jenkins-pipelines/blob/633cb3dd440b6b0d69301e91dd258d764cb0073a/pipelines/jobs/configurations/jdk21u_pipeline_config.groovy#L179
?

Haroon-Khel:
@andrew-m-leonard Right now ive labelled our build-arm-mac machines with `x64` so this will allow the arm mac build jobs to run on these machines. This is only a temp solution to see if we get a stable run of nightlies, and then I can implement a more long term solution using https://github.com/adoptium/ci-jenkins-pipelines/blob/633cb3dd440b6b0d69301e91dd258d764cb0073a/pipelines/jobs/configurations/jdk21u_pipeline_config.groovy#L179

You should know that with this pr merged only build-macstadium-macos11-arm64-1 and build-macstadium-macos11-arm64-2 will be used for building arm and x64 mac for the major jdk versions. I believe testing will still be done both the x64 and arm64 test machines

Haroon-Khel:
I just remembered, for cross compiling jdk11+ the `--openjdk-target=x86_64-apple-darwin` needs to be added to the configure args, so we can merge this until changes are added for this

sxa:
> I just remembered, for cross compiling jdk11+ the `--openjdk-target=x86_64-apple-darwin` needs to be added to the configure args,

Oh https://github.com/adoptium/temurin-build/pull/3492/files was only doing the change for JDK8 - ok in that case I guess we'll bee need another temurin-build PR for 11+.

> we can merge this until changes are added for this

Should that be "can't"?

Haroon-Khel:
>Should that be "can't"?

Havent had my coffee this morning 😅 

>Oh https://github.com/adoptium/temurin-build/pull/3492/files was only doing the change for JDK8 - ok in that case I guess we'll bee need another temurin-build PR for 11+.

On it

Haroon-Khel:
https://github.com/adoptium/temurin-build/pull/3502 is ready to review

andrew-m-leonard (approved):

PR #826, opened 2023-10-10
Enable regular nightly and weekly build for jdk21

sophia-guo:

### DIFF
--- a/pipelines/jobs/configurations/jdk21u.groovy
+++ b/pipelines/jobs/configurations/jdk21u.groovy
@@ -33,9 +33,9 @@ targetConfigurations = [
 ]
 
 // 23:30 Mon, Wed, Fri
-//Uses releaseTrigger_21ea: triggerSchedule_nightly = 'TZ=UTC\n30 23 * * 1,3,5'
+triggerSchedule_nightly = 'TZ=UTC\n30 23 * * 1,3,5'
 // 23:30 Sat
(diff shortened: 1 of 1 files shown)


sophia-guo:
@andrew-m-leonard is it a good time to enable the regular nightly and weekly build for jdk21?

sxa:
I believe the intention is to only use the trigger job going forward - there should be no need to have this running on a more regular cadence, since it will be triggered with all applicable jobs on each new tag appearing.

Also at the moment (while we don't have a JDK21 GA on all platforms) we need to avoid publishing new nightlies for JDK21 so that the EA level that we've been pointing people at is still the one based off jdk-21+35.

andrew-m-leonard:
> @andrew-m-leonard is it a good time to enable the regular nightly and weekly build for jdk21?

Yes, as per @sxa comment.
The jdk21u EA builds will trigger from here https://ci.adoptium.net/job/build-scripts/job/utils/job/releaseTrigger_21ea/
which is currently disabled as per Stewart's comment

sxa:
Having said that I do now have a concern that with no new tags being created in jdk21u perhaps there is justification for having this on some regular basis for jdk21u specifically (maybe just the weekly) but perhaps without publishing

andrew-m-leonard:
Yeah, until Oracle hand over jdk-21.0.2 to SAP in January there will be no regular build tags, and jdk1u will contain a mix of jdk-21.0.1 and 21.0.2 commits

sophia-guo:
Dup of https://github.com/adoptium/ci-jenkins-pipelines/pull/834. Close it.

PR #834, opened 2023-11-02
Build jdk21u HEAD weekly

andrew-m-leonard:
Schedule jdk21u HEAD build at weekend, since no builds are being published upstream until January '24 CPU.


sxa (approved):
SGTM, but noting (as I have elsewhere) that this will override the default nightly/ea retrieved via the API so it is not an equivalent to a tag, however users are able to specify a specific tag (e.g. `jdk-21+35-ea-beta` or `jdk-21.0.1+12-ea-beta` ) as per the examples in [the blog post about ea builds](https://adoptium.net/blog/2023/08/early-access-builds/)

karianna:
> the blog post about ea builds

I wonder if that blog post should be referenced from our nightly builds download pages

PR #836, opened 2023-11-05
Use Docker container for riscv64

luhenry:


luhenry:
It depends on https://github.com/adoptium/infrastructure/pull/3240

karianna (approved):

luhenry:
It is currently blocked by not being able to push the `adoptopenjdk/ubuntu2004_build_image:linux-riscv64` to hub.docker.com. See https://ci.adoptium.net/job/centos7_docker_image_updater/397/console for details

sxa:
This isn't cross compilation  so those options aren't likely to work - this PR is trying to run in an emulated environment so from the build process perspective it would be running "natively". But as you say it isn't going to work yet until we can push it up. Also why have you selected x64 as the host specifically?

luhenry:
> But as you say it isn't going to work yet until we can push it up.

Is there a way to manually push the image until the CI is fixed? Given RISC-V isn't an officially supported platform, the necessity for tracking sources of the tools isn't as much a requirement?

> Why did you select x64 as the host specifically for this?

Kinda of by default. Happy to use something else if it's any better. The other ones I would see is aarch64 if we have enough machines in that pull (and fast enough) to dedicate to that.

sxa (commented):
Can we update the commit message on this - I know the option in the pipelines says "crossCompile" but since this is not using a cross-compiler (it's still running the native riscv64 compiler) I'd prefer it to say "Use system emulation" or something along those lines.

luhenry:
Scaling down the scope of this PR to only using the docker container. We'll see to use x64 host on a follow-up PR.

luhenry:
Superseded by https://github.com/adoptium/ci-jenkins-pipelines/pull/864

PR #840, opened 2023-11-08
Disable jdk17 reproducible comparison builds

sophia-guo:


sxa (commented):
There's no description or issue reference on this PR - why are we disabling it?

andrew-m-leonard:
I realise the jdk17u compares have been red for a while, and we need to have a look at them, but is there another reason to remove them?

sophia-guo:
@jiekang mentioned it's not needed anymore for 17 and we only need to focus on 21.

sxa:
Given the amount of work we've put into backporting stuff into the 17 codebase I'm personally very much -1 on this unless we are actively choose to avoid maintaining it.

sophia-guo:
Based on the discussion we will hold it off for now

PR #843, opened 2023-11-17
update macOS labels to pin XCode version

gdams:
Switches the pinning from OS version to XCode version

Also sets up the labels to work with Orka


Haroon-Khel (approved):

sxa (approved):
Noting here that the jdk11u pipeline is not having a tag applied as it uses the default command line compiler installed by brew, which is a little worrying (Related: https://github.com/adoptium/infrastructure/issues/3249#issuecomment-1815960273)

Also not sure what impact this might have on the corretto builds since we're removing the label, but we don't seem to have built that for close to four years now so 🤷🏻 : https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-corretto/

PR #846, opened 2023-11-21
pin macOS xcode to xcode15.0.1

gdams:
Pinning everything JDK11+ to Xcode 15.0.1


sxa (approved):
Easy to revert if it causes problems so I approve, but let's make sure @andrew-m-leonard has a formal approval too before merging.

PR #853, opened 2023-12-06
Change default activeNodeTimeout from 0 to 1 min

smlambert:
Addresses the 30 second lag for spinning up Orka machines, but also seems to be a more sensible default timeout setting than zero minutes (especially since it doesn't delay sending jobs to machines if they are available in zero minutes).

Related: https://github.com/adoptium/aqa-tests/issues/4899


andrew-m-leonard (approved):
lgtm

zijwij-byxmyq-1wIxre (commented):
Add

Issue #847, opened 2023-11-22
Create new testJobType='release'


smlambert:
Currently, we support 2 testJobType values, 'nightly' and 'weekly' and then for releases, we combine 'nightly' + 'weekly' test lists to create the list that is run during a release.  This limits our ability to add new tests that we want to try out in weekly runs, but do not want to include during a release run.  To give us more flexibility, adding a new testJobType called 'release' will be the simplest way forward.

There are a few places in the code that need updating, including:
https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/build_base_file.groovy#L254-L310
https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/defaults.json#L42-L54

PR #854, opened 2023-12-11
Use parameter releaseType to pass to downstream

AdamBrousseau:
Rather than hardcoding 'Weekly', use the build parameter.

### DIFF
--- a/pipelines/build/common/weekly_release_pipeline.groovy
+++ b/pipelines/build/common/weekly_release_pipeline.groovy
@@ -29,7 +29,7 @@ stage('Submit Release Pipelines') {
         cleanWs notFailBuild: false
     }
 
-    // For each variant, create a weekly pipeline job with "Weekly" as "releaseType"
+    // For each variant, launch a pipeline job with "releaseType" based on the build parameter.
     scmRefs.each { variant ->
         def variantName = variant.key
         def scmRef = variant.value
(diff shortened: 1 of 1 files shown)


smlambert:
> If this is the weekly config is it not OK to hard code that?

In https://github.com/adoptium/ci-jenkins-pipelines/issues/847 we have introduced a new job type for releases, so now there are "nightly", "weekly" and "release" types.  The name of the file is now a misnomer, weekly_release_pipeline can be used for weekly runs or release runs and it can take different job types now.

AdamBrousseau:
>If this is the weekly config is it not OK to hard code that?

It is already hardcoded in the build_pipeline_generator.groovy
https://github.com/adoptium/ci-jenkins-pipelines/blob/833a06de53ae21ffec043863600a88a5dbbb6529/pipelines/build/regeneration/build_pipeline_generator.groovy#L265

This sets the default param value for the weekly pipelines to `releaseType=Weekly`. If a build is launched from the weekly pipeline and releaseType param is changed, it has no effect. Perhaps by design? But a little misleading I think.

adamfarley (changes requested):
The code looks good. Can we amend the comment about "create a weekly pipeline job" to reflect this change, please?

sxa (commented):
Can you redo the commit message to include the `e` in the middle of `releaseType` please?

karianna:
> > If this is the weekly config is it not OK to hard code that?
> 
> In #847 we have introduced a new job type for releases, so now there are "nightly", "weekly" and "release" types. The name of the file is now a misnomer, weekly_release_pipeline can be used for weekly runs or release runs and it can take different job types now.

We should probably rename weekly config files to what they really are in that case :-)

AdamBrousseau:
I'll further explain my use case.
We (IBM) do not publish our nightly/weekly build. So we've always run "nightly without publish" and "weekly without publish". I've reconfigured our regenration code to make the defaults "* without publish". I recently made a change to properly set the config rather than just reordering the drop down list in the build launch (see https://github.com/ibmruntimes/ci-jenkins-pipelines/pull/188). In doing so, we're relying on the parameter passed from the weekly build to the regular build, which up until this PR, has been ignored because it takes the hardcoded "Weekly". So, we are not reinventing the weekly build (wrapper job) into something else, we are simply using it "without publish". Arguably, the publish should be a separate boolean and not part of the release type. But that is a different topic.

smlambert:
For the record, I agree with you on the cleaner design of separate boolean param for 'publish', but as you say, its outside of the scope of this PR.

adamfarley (approved):
LGTM, thanks for changing the comment. :)

Issue #855, opened 2023-12-12
Fix Labeller config file to new v5 syntax


karianna:
Somewhere along the lines we moved to v5 which has totally new syntax https://github.com/actions/labeler/issues/712

sxa:
Presumably this is the cause of the labeler failures on every PR at the moment?
```
The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
Error: Error: found unexpected type for label 'ghActions' (should be array of config options)
Error: found unexpected type for label 'ghActions' (should be array of config options)
```

karianna:
#872 attempts to fix this

PR #860, opened 2023-12-15
fix: use `actions/labeler@v5` config file format

lemeurherve:
This PR adapts the labeler config file to the new format introduced in version 5.

Related to:
- #855

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,30 +5,40 @@
 # e.g to add the test label to any changes in the /tests directory:
 
 # test:
+# - changed-files:
 #   - tests/*
 
 ghActions:
-    - .github/workflows/**/*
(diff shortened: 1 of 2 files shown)


lemeurherve:
I signed-off my commit with the mail I've used to create my Eclipse account and sign the ECA, I don't know what to do more.

karianna:
@lemeurherve I revalidated the signing process so that passes now.  Reading the labeler documentation tells me your change is likely correct but it's hard to tell since pull_request_target only runs against the base branch.  I'll merge this as it's a step in the right direction.

karianna (approved):
This is a step in the right direction

karianna:
Needs one more approval

smlambert (approved):

PR #864, opened 2023-12-20
Use dockerhost-rise-ubuntu2204-aarch64-1 host to run in QEMU for riscv64

luhenry:


sxa (approved):
Probably not needed for pre-JDK21 but it won't do any harm. I /think/ this will work, but I'm happy to approve and we can look at fixing up later if there are any problems.

luhenry:
@sxa how can we get merged?

sxa:
> @sxa how can we get it merged?

As per the message from github it needs two reviewers. I have approved it already, but you need a second.

smlambert (approved):
👍

sxa (approved):

sxa (changes requested):
Oh hang on - just noticed before merging that there are linter failures - those will need to be addressed first
EDIT: It's missing the comma at the end of the `crossCompile` line.

pipelines/jobs/configurations/jdk21u_pipeline_config.groovy
```diff
@@ -166,6 +166,9 @@ class Config21 {
         riscv64Linux      :  [
                 os                  : 'linux',
                 arch                : 'riscv64',
+                // crossCompile        : 'dockerhost-rise-ubuntu2204-aarch64-1',
```
luhenry, line 169: @sxa can we actually enable this one as well?

pipelines/jobs/configurations/jdk22_pipeline_config.groovy
```diff
@@ -145,6 +145,9 @@ class Config22 {
         riscv64Linux      :  [
                 os                  : 'linux',
                 arch                : 'riscv64',
+                // crossCompile        : 'dockerhost-rise-ubuntu2204-aarch64-1',
```
luhenry, line 148: @sxa same as jdk21u

sxa:
Yeah once we get the boot JDK in then we can definitely add the other two pipelines.
I'm not quite sure why the `Test / Groovy` check is failing just now though ...

pipelines/jobs/configurations/jdk21u_pipeline_config.groovy
sxa: Since we're using 21 to bootstrap 21 on RISC-V then yeah as soon as we've got a suitable boot JDK that can be used then we can enable this.
On my "static" containerized build environemnts on my machines I've manually installed the [21.0.1+12.1 JDK](https://api.adoptium.net/v3/binary/version/jdk-21.0.1+12.1-ea-beta/linux/riscv64/jdk/hotspot/normal/adoptium) and symlinked to `/usr/lib/jvm/jdk-21` - we should set up the playbooks to do that.

sxa (approved):
I'm going to merge this for 17 just now so we can get it tested in production - we can look at enabling in a subsequent PR for 21+ (Noting that JDK23 will fail until we fix the boot JDK download issue)

PR #871, opened 2024-01-02
Fix Groovy failing build and update dependencies within major versions

karianna:

PR #872, opened 2024-01-02
2nd attempt to fix v5 labeler

karianna:
WARN: Because of https://github.com/actions/labeler?tab=readme-ov-file#notes-regarding-pull_request_target-event this PR will appear to have failed the GH Action check.  We won't know its worked until a subsequent PR is created.  This is why we can't have nice things :-)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -10,35 +10,28 @@
 
 ghActions:
 - changed-files:
-  - .github/workflows/**/*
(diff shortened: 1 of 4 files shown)


karianna:
https://github.com/adoptium/ci-jenkins-pipelines/pull/871 fixes the groovy issue separately

smlambert (approved):

andrew-m-leonard (approved):

PR #873, opened 2024-01-04
Sanitise parameters

sxa:
This PR will trap potentially dangerous characters before passing them to a shell operation within the groovy pipelines. Noting that some dodgy characters are trapped elsewhere (but doing it here makes it explicit in case other places are missed):
```
15:45:46  Execution error: ERROR: Name must follow the pattern &#039;^[a-zA-Z0-9]+((\.|_|__|-+)[a-zA-Z0-9]+)*$&#039;
[Pipeline] echo
15:45:46  ERROR: Name must follow the pattern &#039;^[a-zA-Z0-9]+((\.|_|__|-+)[a-zA-Z0-9]+)*$&#039;
```
If the docker image parameter has nonsense in it.
Also VARIANT will get trapped and abort in make-adopt-build-farm.sh if it's not in a defined list: `16:29:19  + ./build-farm/make-adopt-build-farm.sh
16:29:19  [ERROR] hot$pot is not a recognised build variant. Valid Variants = hotspot temurin openj9 corretto SapMachine dragonwell fast_startup bisheng`

Ref TOB_TEMURIN_17

Preferred future solution would be to avoid using `sh` where possible to reduce the risk of shell escapes, but this should be good as an interim solution.

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -906,6 +906,11 @@ class Builder implements Serializable {
                                         //Remove the previous artifacts
                                         try {
                                             context.timeout(time: pipelineTimeouts.REMOVE_ARTIFACTS_TIMEOUT, unit: 'HOURS') {
+                                                if ( ! ( "${config.TARGET_OS}"    ==~ /^[A-Za-z0-9\/\.-_]$/ ) ||
+                                                     ! ( "${config.ARCHITECTURE}" ==~ /^[A-Za-z0-9\/\.-_]$/ ) ||
+                                                     ! ( "${config.VARIANT}"      ==~ /^[A-Za-z0-9\/\.-_]$/ ) ) {
+                                                    throw new Exception("[ERROR] Dubious character in TARGET_OS, ARCHITECTURE or VARIANT - aborting");
+                                                }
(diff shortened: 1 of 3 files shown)


karianna:
@sxa Can you rebase? That should fix CI checks

sxa:
> @sxa Can you rebase? That should fix CI checks

I've puilled in your two commits from pending PR https://github.com/adoptium/ci-jenkins-pipelines/pull/871/ and that seems to have [allowed the checks to pass](https://github.com/adoptium/ci-jenkins-pipelines/actions/runs/7420874211/job/20193071034?pr=873). Since they're not yet merged upstream I'm now going to remove them from this PR again, but that should be enough to prove that this PR is good.

karianna (approved):

steelhead31 (approved):
LGTM

AdamBrousseau:
I'm caught up on the aqa branch regex. We have an aqa release branch named `j9.0.42-release` for a release build that just got caught on this. I started looking at the regex but I'm failing to see how the it works for any of the [aqa branches](https://github.com/adoptium/aqa-tests/branches). Unless I'm misunderstanding or missing something. I also tried `master` and `v0.9.9-release` and both fail.

PR #879, opened 2024-01-09
Add rerunIterations

### LINKED ISSUE #777: Trial the 'automatic rerun on failure' feature for test jobs
Test jobs have an 'automatic rerun on failure' feature (https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L1216) which we should try using in our regular runs.

That feature is turned on via the RERUN_ITERATIONS parameter.  So in order to have it enabled, we can set that parameter in the jobs we trigger from the build pipeline (see https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L445-L462).

As a trial, we will minimally set RERUN_ITERATIONS=1 (if greater than 1, will launch in parallel, but let's not do that yet, but rather see if a simple rerun helps resolve intermittent failures we sometimes see).

smlambert:
Fixes #777 

Retry running certain test jobs automatically if they fail (wasteful if applied to unstable jobs that do not have up-to-date exclusions and triage done against them).

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -371,6 +371,10 @@ class Build {
                     context.println "Running test: ${testType}"
                     context.stage("${testType}") {
                         def keep_test_reportdir = buildConfig.KEEP_TEST_REPORTDIR
+                        def rerunIterations = '3'
+                        if ("${testType}".contains('dev') || "${testType}".contains('external')) {
+                            rerunIterations = '0'
+                        }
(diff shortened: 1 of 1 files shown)


karianna (approved):

steelhead31 (approved):
LGTM

sxa (approved):
Since this has been trialled with Semuru I'm going to assume this is safe so giving approval. A couple of notes:
1. How does TRSS handle the re-runs?
2. Since this is going to require regeneration for the new parameter it will presumably exacerbate https://github.com/adoptium/infrastructure/issues/2774 in the short term until we get every job we have regenerated

PR #881, opened 2024-01-09
Allow : in build docker image name

sxa:
Also escapes the `-` character and adjusts the formatting of the error message. This was causing a problem for the riscv64 which use a slightly different format for these values and is causing a build job failure after the merging of https://github.com/adoptium/ci-jenkins-pipelines/pull/873

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1885,9 +1885,9 @@ class Build {
 
 
                         context.println "[NODE SHIFT] MOVING INTO DOCKER NODE MATCHING LABELNAME ${label}..."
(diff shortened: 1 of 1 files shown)


Haroon-Khel (approved):

smlambert:
In the long-run, is there a way to establish an actual naming convention and have riscv follow it, rather than having to make concessions for every case that doesn't follow a convention?

andrew-m-leonard (approved):

sxa:
> In the long-run, is there a way to establish an actual naming convention and have riscv follow it, rather than having to make concessions for every case that doesn't follow a convention?

We've never established an explicit convention but what RISC-V is as close as it reasonably be to everything else so is absolutely fine with a tag with the platform name in it and a parameter to tell docker that it's using a non-native container (for performance reasons) which is why it's different. It's a case of trailblazing and needing to have special parameters to do that. The parameters are broadly consistent with what Alibaba did for test.

Yeah we could make the convention more fixed, but it would have to include what we've done in this situation.

PR #888, opened 2024-01-12
Fix missing wildcard in os/arch/variant pattern match

sxa:
Fixes #887, introduced by #873
Will need to be cherry picked across to the v2024.01.01 branch once approved

PR #891, opened 2024-01-12
Escape '-' literals in patterns

keithc-ca:
The pattern `[.-_]` matches characters `0x2e` through `0x5f` (which includes `[A-Z]`). I don't think that was the intent; `-` must be escaped (or appear first or last in `[]`).

See also #881 and #888.


sxa (approved):
Thanks Keith - can you PR the same change against the v2024.01.01 branch too please?

smlambert (approved):

keithc-ca:
> can you PR the same change against the v2024.01.01 branch too please?

See https://github.com/adoptium/ci-jenkins-pipelines/pull/893.

PR #910, opened 2024-02-01
Moving prepare stage to node 'worker' instead of 'Linux/x64'

### LINKED ISSUE #906: reproducible comparison jobs for other platforms are taking up Linux/x64 nodes
**What are you trying to do?** Use the dockerBuild linux executors (or in today's case, mark the host offline for temporary maintenance)

**Expected behaviour:** Executors are available

**Observed behaviour:** Executors are tied up with reproduce_compare jobs for other platforms as per the following screenshot:
image
It appears that we require a node with label `dockerBuild&&linux&&x64` regardless of which reproducible build platform we are running.

**Any other comments:**
While https://github.com/adoptium/ci-jenkins-pipelines/issues/905 covers a potential deadlock where the reproduce_compare jobs for xLinux are deadlocking, this is a separate issue to cover discussion around the use of xlinux executors for other platforms, placing unnecessary load on those systems at a time when we are reducing capacity as a result of the Equinix removal.

Also in the current situation the reproducible compare jobs are continuing even when their parent pipeline is complete, so we're in a situation where [mac compare 96](https://ci.adoptium.net/job/jdk17u-mac-x64-temurin_reproduce_compare/96/console) from the screenshot above is still waiting, but it's [parent pipeline](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk17u/job/jdk17u-mac-x64-temurin/485/console) has ended. A similar situation is happening for Windows where the [windows compare job](https://ci.adoptium.net/job/jdk21u-windows-x64-temurin_reproduce_compare/30/) was initiated by the aborted parent https://github.com/adoptium/ci-jenkins-pipelines/pull/904. Similarly for Linux/s390x the [compare job](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk17u/job/jdk17u-linux-s390x-temurin/411/) was initiated by [parent job](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk17u/job/jdk17u-linux-s390x-temurin/411/) which has completed. In all three cases another subsequent pipeline is running, which blocks the ability of the compare job to run until that new one is complete.

Noting that for each of these situations, the parent job was in a different state (Aborted for Windows, Failed for mac, warnings for Linux/s390x) so that does not seem to affect the behaviour)

sophia-guo:
Moving prepare stage to node 'worker' instead of 'Linux/x64', which is has more abundant executors than 'linux/x64'.

Close #906

### DIFF
--- a/tools/reproduce_comparison/Jenkinsfile
+++ b/tools/reproduce_comparison/Jenkinsfile
@@ -3,7 +3,7 @@ Jenkins job does reproducible build compare.
 */
 import groovy.json.JsonOutput
 
-env.NODE_LABEL = 'dockerBuild&&linux&&x64'
+env.NODE_LABEL = 'worker'
 pipeline {
     agent none
     parameters {


sxa (approved):
Thanks :-)

karianna (approved):

PR #920, opened 2024-02-09
Make sure betaTrigger job does not lock worker node

andrew-m-leonard:
betaTrigger job locks the worker node while the build is running, which it should not do
Moved parallel jobs out of worker node context

Test: https://ci.adoptium.net/job/build-scripts/job/utils/job/betaTrigger_8ea/26/console


sxa (approved):
On the basis that Andrew has confirmed that this should just avoid locking any node, and not lock any random alternate node I'm ok with this

steelhead31 (approved):
LGTM

PR #921, opened 2024-02-09
Allow for . in beta binaries status check of archive EA filename

andrew-m-leonard:
New beta EA filename has "." in the version part of the filename, the nightly_build_and_test_stats.groovy regex needs updating to match it correctly.

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -150,7 +150,7 @@ def verifyReleaseContent(String version, String release, String variant, Map sta
                     // Search for artifacts in the releaseAssets list
                     ftypes.each { ftype ->
                         def arch_fname = archToAsset[osarch]
-                        def findAsset = releaseAssets =~/.*${file_image}_${arch_fname}_[^\."]*${ftype}".*/
+                        def findAsset = releaseAssets =~/.*${file_image}_${arch_fname}_[^"]*${ftype}".*/
                         if (!findAsset) {
                             missingForArch.add("$osarch : $image : $ftype".replaceAll("\\\\", ""))
                         } else {


karianna (approved):

sxa (approved):
I'm a bit nervous as to why `\.` was in there in the first place in case this breaks something, but I guess it's easily fixable if it causes a problem so I'm approving.

Noting that the old versions had a name like this:

- [OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21-0-1-12.tar.gz](https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.1%2B12-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21-0-1-12.tar.gz)

And the new one is:
- [OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21.0.3-0-1.tar.gz](https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.3%2B1-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21.0.3-0-1.tar.gz)

See https://github.com/adoptium/temurin-build/issues/3355#issuecomment-1568129358 for the reasons why it was done with hyphens originally - it was to do the minimal amount of work needed to avoid changing the existing release tool job.

andrew-m-leonard:
> I'm a bit nervous as to why `\.` was in there in the first place in case this breaks something, but I guess it's easily fixable if it causes a problem so I'm approving.
> 
> Noting that the old versions had a name like this:
> 
>     * [OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21-0-1-12.tar.gz](https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.1%2B12-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21-0-1-12.tar.gz)
> 
> 
> And the new one is:
> 
>     * [OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21.0.3-0-1.tar.gz](https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.3%2B1-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21.0.3-0-1.tar.gz)
> 
> 
> See [adoptium/temurin-build#3355 (comment)](https://github.com/adoptium/temurin-build/issues/3355#issuecomment-1568129358) for the reasons why it was done with hyphens originally - it was to do the minimal amount of work needed to avoid changing the existing release tool job.

I think basically I over did it last time !

PR #923, opened 2024-02-12
Trigger the release dry run right after each jdk build is done

### LINKED ISSUE #911: Auto build the dry run release in post build stage
The current release publish workflow is 

1.  dry run - manually set the parameter of job https://ci.adoptium.net/job/build-scripts/job/release/job/refactor_openjdk_release_tool/
2. rerun the dry run job with dry run parameter unchecked.

The mistakes might happen when manually setting the parameters besides sometime it's hard to find the dry run build if the job is triggered by nightly build automatically at the same time.

If the dry run can be triggered automatically by the pipeline build the process  would be easier and more reliable  for  for release champions.

sophia-guo:
Trigger the release dry run right after each jdk build is done
Provide the release publish build link per os, arch with pre-filled parameters

Close #911

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -770,11 +770,6 @@ class Builder implements Serializable {
     Call job to push artifacts to github. Usually it's only executed on a nightly build
     */
     def publishBinary() {
-        if (release) {
-            // make sure to skip on release
-            context.println('Not publishing release')
-            return
-        }
 
         def timestamp = new Date().format('yyyy-MM-dd-HH-mm', TimeZone.getTimeZone('UTC'))
         def tag = "${javaToBuild}-${timestamp}"
@@ -784,8 +779,8 @@ class Builder implements Serializable {
         }
 
         context.stage('publish') {
-        context.println "publishing with publishName: ${publishName}"
-        context.build job: 'build-scripts/release/refactor_openjdk_release_tool',
+            context.println "publishing with publishName: ${publishName}"
+            context.build job: 'build-scripts/release/refactor_openjdk_release_tool',
                     parameters: [
                         ['$class': 'BooleanParameterValue', name: 'RELEASE', value: release],
                         ['$class': 'BooleanParameterValue', name: 'DRY_RUN', value: false],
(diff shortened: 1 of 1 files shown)


sophia-guo:
The linter warning is the methods is 31, the default maxMethods=30 https://codenarc.org/codenarc-rules-size.html#methodcount-rule

karianna (approved):

sxa (approved):
As I understand it this will happen as part of the top level openjdkXX-pipeline which sounds good to me!

sophia-guo:
> As I understand it this will happen as part of the top level openjdkXX-pipeline which sounds good to me!

Correct. Start the dry run from the top level openjdkXX-pipeline

sxa:
I didn't spot this at first but it looks like this failed the [github linter check](https://github.com/adoptium/ci-jenkins-pipelines/actions/runs/7889769348/job/21530405964?pr=923) - possibly as a result of it going over the linter's limit for the number of methods in a class with new ones being added here:
```
  34    warning  Class Builder has 31 methods  MethodCount             
```
I can't see anything else that was flagged in there, but this will presumably cause subsequent PRs to the pipelines to fail the checks too.

sophia-guo:
https://github.com/adoptium/ci-jenkins-pipelines/pull/923#issuecomment-1939720738 @sxa

sophia-guo:
> Eventually I could see dryrunReleasePublish(IndividualBuildConfig config) becoming publishRelease(IndividualBuildConfig config, BooleanParameterValue dryRun)

We can do that when the betaTrigger is finalized.

sxa:
> The linter warning is the methods is 31, the default maxMethods=30 https://codenarc.org/codenarc-rules-size.html#methodcount-rule

Yeah but we need to decide what to do with it since at the moment it's going to cause check failures which we don't want - either we need to reduce the number of methods or adjust the linter configuration to avoid problems for contributors. I'll raise an issue.

sophia-guo:
https://github.com/adoptium/ci-jenkins-pipelines/issues/926 can fix this temporarily, which downsize the method number to 30. But eventually either adjust the linter configuration or redefining the methods of builder is needed. Can open a separate issue @sxa

Issue #926, opened 2024-02-13
Combine publishBinary() and dryrunReleasePublish(IndividualBuildConfig config)


sophia-guo:
Combine publishBinary() and dryrunReleasePublish(IndividualBuildConfig config) to publishRelease(BuildConfigs configs, BooleanParameterValue dryRun)

Note that publishBinary() is called when all jdk builds are done. dryrunReleasePublish() is called per each jdk build is done.

PR #916, opened 2024-02-06
Trigger the release dry run

sophia-guo:
Close #911

PR #931, opened 2024-02-15
postBuild plugin manager only available in postBuild groovy

andrew-m-leonard:
pipeline builds broken as postBuild plugin "manager" is only available in postBuild groovy scripts
https://ci.adoptium.net/job/build-scripts/job/openjdk17-pipeline/899/console
```
hudson.remoting.ProxyException: groovy.lang.MissingPropertyException: No such property: manager for class: Builder
```

I've commented out the build summary annotation for now

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -824,6 +824,7 @@ class Builder implements Serializable {
                 releaseToolUrl += "&DRY_RUN=true"
             }
             URLEncoder.encode(releaseToolUrl.toString(), "UTF-8")
+            echo "dryrunReleasePublish: ${releaseComment} : ${releaseToolUrl}"
             return ["${releaseToolUrl}", "${releaseComment}"]
         }
     }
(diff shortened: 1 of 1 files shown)


smlambert (approved):

sxa (commented):
Just to be clear - this is a partial revert of https://github.com/adoptium/ci-jenkins-pipelines/pull/923 if I'm reading it right - it would be good to note which functionality will be lost from that original PR when this is merged. I suggest @sophia-guo reviews it too to ensure she is aware.

andrew-m-leonard:
> Just to be clear - this is a partial revert of #923 if I'm reading it right - it would be good to note which functionality will be lost from that original PR when this is merged. I suggest @sophia-guo reviews it too to ensure she is aware.

Correct, it's just the build summary label text that will be missing, instead i've put an "echo" with the same information

sxa (approved):
Approving so this can go in for tonight's builds - if @sophia-guo has any objections we can address them afterwards, but this is causing a build break.

sophia-guo:
I will create a PR to fix this. As the echo only shows the message of link in console output instead of a hyperlink in pipeline job description.

sophia-guo:
Should be a simple update manager --> context.manager.

sophia-guo:
I have tested the update `manager --> context.manager.` works. I will do the update together with the PR for #926

andrew-m-leonard:
> I have tested the update `manager --> context.manager.` works. I will do the update together with the PR for #926

@sophia-guo ah good.
Another thing i've noticed from the overnight beta builds, is it's being called for non-release builds which is probably fine as provides a good test of it, but it's failing as non-release publishes need a TIMESTAMP field: https://ci.adoptium.net/job/build-scripts/job/release/job/refactor_openjdk_release_tool/8236/

sophia-guo:
@andrew-m-leonard interesting. Ah, switch between https://github.com/adoptium/ci-jenkins-pipelines/pull/923/files and https://github.com/adoptium/ci-jenkins-pipelines/pull/916, forgot the condition if (release). The release_tool build was called  platform first and then after all jdks are done release_tool build was called again (this is the final one and correct one). This failed release_tool build per platform but won't affect the final one. No worries about the publish but confused and also waste build resources. This will also be fixed by https://github.com/adoptium/ci-jenkins-pipelines/pull/932

PR #934, opened 2024-02-19
Ensure beta EA builds triggered during release periods do not enable Testing

andrew-m-leonard:
During release periods:
**previous Saturday to the following Sunday around the release Tuesday of Jan, March, Apr, Jul, Sep, Oct**
do not enable Testing for triggered beta EA builds

Upstream release day is the closest Tuesday to the 17th of the month.

### DIFF
--- a/pipelines/build/common/trigger_beta_build.groovy
+++ b/pipelines/build/common/trigger_beta_build.groovy
@@ -47,6 +47,7 @@ def isDuringReleasePeriod() {
     def releasePeriod = false
     def now = ZonedDateTime.now(ZoneId.of('UTC'))
     def month = now.getMonth()
+month = Month.MARCH
 
     // Is it a release month?
     if (month == Month.JANUARY || month == Month.MARCH || month == Month.APRIL || month == Month.JULY || month == Month.SEPTEMBER || month == Month.OCTOBER) {
--- a/pipelines/build/common/trigger_beta_build.groovy
+++ b/pipelines/build/common/trigger_beta_build.groovy
@@ -89,7 +89,7 @@ node('worker') {
     def binariesRepoTag = publishJobTag + "-beta"
 
     if (isDuringReleasePeriod()) {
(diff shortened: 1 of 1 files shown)


sxa:
My honest view is that this is adding quite a bot of complexity in terms of the code for marginal benefit over our current processes. If we only disabled TESTING then I'd be ok with this (which is our current policy) but I personally don't agree with shutting off the builds during the release cycle, especially for people who are looking to trial the latest tip builds on a regular basis.

andrew-m-leonard:
> My honest view is that this is adding quite a bot of complexity in terms of the code for marginal benefit over our current processes. If we only disabled TESTING then I'd be ok with this (which is our current policy) but I personally don't agree with shutting off the builds during the release cycle, especially for people who are looking to trial the latest tip builds on a regular basis.

Yes, I think I agree, changing to disable testing is probably best.

My concern is the possibility of triggering an ea-beta on a non "-ga" tagged release build, but to that to happen the release tag would have to have been published and the "-ga" delayed, and that to happen over the 1am beta trigger period. I think the slim possibility is probably not too much to be concerned about.

I'll change to disable testing...

sxa:
Thanks :-)

Just noting that for a sample release (January's jdk17u):
- jdk-17.0.10+7 was tagged on a commit from January 9th so (if it was tagged at that time and not just before GA) then it would already have triggered and built without interfering the release. If both tags were created on release day then there is the - hopefully small - window you describe if it happened around 1am.
- The next tag - jdk-17.0.11+1 was on January 31st which would hopefully have been after all of the build and test was completed so again wouldn't have interfered with them

The changes in this PR would have had no effect on anything around that period, but to cover all bases I think this change is ok. I suppose there is still some potential for taking machine resource away from our pre-GA dry-run pipelines but we can see how that works out

I'll also note that we could still run the full set of testing on the evaluation pipelines since they don't share machine resource with any release platforms at the moment, but I'm ok with trimming the testing on those during the period too if it makes life harder to retain it.

karianna (approved):

PR #947, opened 2024-02-29
Update linux-riscv64 jdk11u pipeline to temurin

luhenry:
The backport has been merged upstream with [1].

[1] https://github.com/openjdk/riscv-port-jdk11u/commit/309291f166821443393818beed364de6f2de5c52

### DIFF
--- a/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
@@ -205,7 +205,7 @@ class Config11 {
                 ],
                 configureArgs        : [
                         'hotspot'    : '--enable-headless-only=yes --enable-dtrace',
-                        'temurin'    : '--enable-headless-only=yes --enable-dtrace',
+                        'temurin'    : '--enable-headless-only=yes --enable-dtrace --disable-ccache',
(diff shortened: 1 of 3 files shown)


luhenry:
Depends on https://github.com/adoptium/temurin-build/issues/3677

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
```diff
@@ -183,29 +183,35 @@ class Config11 {
...
                 ],
                 crossCompile         : [
                         'hotspot'    : 'dockerhost-rise-ubuntu2204-aarch64-1',
+                        'temurin'    : 'dockerhost-rise-ubuntu2204-aarch64-1',
                         'openj9'     : 'x64',
                         'bisheng'    : 'x64'
                 ],
                 buildArgs            : [
                         'hotspot'    : '--create-sbom',
+                        'temurin'    : '--create-sbom',
```
sxa, line 202: Should we remove the hotspot line from these while replacing with Temurin?

sxa (approved):
Just one thought but not critical that it is addressed

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
luhenry: I wasn't sure what the policy should be, happy to go either way

luhenry: I will remove then

PR #962, opened 2024-03-11
Fix Signverify Job - Change WindowsKitPath to find latest signtool rather than earliest/first

### LINKED ISSUE #917: Windows aarch64 sign_verification not running signtool.exe correctly
https://ci.adoptium.net/job/build-scripts/job/release/job/sign_verification/816/console
```
C:/Users/jenkins/workspace/build-scripts/release/sign_verification@tmp/durable-a56db375/script.sh: line 12: /cygdrive/c/Program Files (x86)/Windows Kits/10/bin/10.0.16299.0/arm64/signtool.exe: Accessing a corrupted shared library
Error: executable not Signed: unpacked/jre/jdk-21.0.3+1-jre/bin/j2pcsc.dll
```

signing_verify job has to find the signtool.exe it then uses to verify the exe/dlls, on cross compiled aarch64 it looks like it's finding the wrong one...?

steelhead31:
Fixes #917 

Amend the search for signtool.exe to perform a reverse sort on the search for signtool.exe. 

On arm64 in particular, the earliest version ( 17763 ) does not actually work, which I believe is a bug in the earlier SDKS, and the later versions work correctly.

### DIFF
--- a/pipelines/build/common/verify_signing.groovy
+++ b/pipelines/build/common/verify_signing.groovy
@@ -40,7 +40,7 @@ String find_signtool() {
 
     def windowsKitPath = "/cygdrive/c/'Program Files (x86)'/'Windows Kits'"
 
-    def files = sh(script:"find ${windowsKitPath} -type f -path */${arch}/signtool.exe", \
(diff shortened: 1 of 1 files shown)


andrew-m-leonard (approved):
lgtm

PR #964, opened 2024-03-11
Prevent Mac sign verification from running on old OS versions

adamfarley:
Because later versions of OpenJDK (e.g. JDK22) cannot run on MacOS versions below 11. This extra tag prevents us from attempting to do so.

Making it universal rather than version-specific because I don't see any benefit from sign verification on MacOS 10.14, specifically. Testing yes, sign verification no.


smlambert (approved):

steelhead31 (approved):
LGTM

gdams (approved):

PR #972, opened 2024-03-15
Define releaseSummary for publish and release

sophia-guo:


andrew-m-leonard (approved):
lgtm

andrew-m-leonard:
Thanks Sophia, so yeah this also fixes the fact that the currentBuild.setKeepLog & displayName() was not occuring, thanks

smlambert (approved):

Issue #981, opened 2024-03-21
Issue when running build pipeline which do not have sign_temurin_gpg


mahdipub:
- no jenkins pipeline job
- all platforms

I have a question regarding a piece of build code.
[/pipelines/build/common/openjdk_build_pipeline.groovy](https://github.com/adoptium/ci-jenkins-pipelines/blob/d966e612efa09a83175debe08fa187c2d07e9c2c/pipelines/build/common/openjdk_build_pipeline.groovy#L2128)
```
if (!env.JOB_NAME.contains('pr-tester')) {
    try {
        gpgSign()
    } catch (Exception e) {
        context.println(e.message)
    }
}
```
This condition always is true in our environment for all platforms and we will end up failure as we do not have any `gpgSign() -> build-scripts/release/sign_temurin_gpg` to run. 

Can we have a more accurate condition here to avoid it run for non-adoptium users?

AdamBrousseau:
I think the ask is if we can submit a change to add a condition to that `if` line. Either `!openj9` or `JENKINS_URL.contains('adopt')` like we did here
https://github.com/ibmruntimes/ci-jenkins-pipelines/blob/3d3cc13c466878349f5d6702381c6def555d4250/pipelines/build/common/create_job_from_template.groovy#L71

AdamBrousseau:
I see in the Adopt OpenJ9 builds the sign_temurin_gpg job is run
ex
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-ppc64le-openj9/1339/consoleFull
```
14:42:36  RUNNING sign_temurin_gpg for linux/ppc64le ...
[Pipeline] build (Building build-scripts » release » sign_temurin_gpg)
14:42:36  Scheduling project: [build-scripts » release » sign_temurin_gpg](https://ci.adoptium.net/job/build-scripts/job/release/job/sign_temurin_gpg/)
14:42:45  Starting building: [build-scripts » release » sign_temurin_gpg #27185](https://ci.adoptium.net/job/build-scripts/job/release/job/sign_temurin_gpg/27185/)
14:44:47  Build [build-scripts » release » sign_temurin_gpg #27185](https://ci.adoptium.net/job/build-scripts/job/release/job/sign_temurin_gpg/27185/) completed: SUCCESS
[Pipeline] node
[...]

PR #983, opened 2024-03-27
avoid sign_temurin_gpg in non-adopt pipelines

mahdipub:
fixes issue #981  

This is to avoid not adoptium pipelines pick the sign_temurin_gpg

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -2125,7 +2125,7 @@ class Build {
                         throw new Exception("[ERROR] Installer job timeout (${buildTimeouts.INSTALLER_JOBS_TIMEOUT} HOURS) has been reached OR the downstream installer job failed. Exiting...")
                     }
                 }
-                if (!env.JOB_NAME.contains('pr-tester')) {
+                if (!env.JOB_NAME.contains('pr-tester') && JENKINS_URL.contains('adopt')) {
                     try {
                         gpgSign()
                     } catch (Exception e) {


smlambert (approved):
Approving it because I assume no downstream user of these pipelines would want this block, though I have not confirmed with anyone about that assumption.

sxa (commented):
I'm going to defer to @andrew-m-leonard on this. While I'd be ok with it (and it's something I've done elsewhere) I feel there ought to be a parameter that can be used to determine whether to invoke the signing code.

@AdamBrousseau has suggested a `VARIANT` check as potentially suitable for this in https://github.com/adoptium/ci-jenkins-pipelines/issues/981#issuecomment-2013789159 - I'd be ok with it being `if VARIANT = temurin` as a preference if that's feasible since I don't think we'll ever have a requirement to be doing this on another variant (since Adoptium don't ship anything else)

AdamBrousseau:
There is `ENABLE_SIGNER` param. But we use that to toggle our own signing process. We kick a different job to sign our stuff with presumably different params. If you don't care to sign your OpenJ9 builds we could use variant=temurin here. I figured since you are currently signing your openj9 builds we'd go this route.

andrew-m-leonard (approved):
Although I feel a new param maybe ideal, we do have precedence for this same check on JENKINS_URL for things like pipeline generation authentication...
So I am good with this.

sxa (approved):
> Although I feel a new param maybe ideal, we do have precedence for this same check on JENKINS_URL for things like pipeline generation authentication... So I am good with this.

👍🏻 in which case I'll approve too. It's a terrible (IMHO) precedent, and perhaps we should look at going through them all later (I hadn't realised it was in other places too) but I'm ok with this extra 'leak' for now.

PR #984, opened 2024-03-28
Enable Temurin build of x64/ppc64le/aarch64/s390x Linux for jdk-21+ using Adoptium gcc 11.3.0 DevKit

andrew-m-leonard:
Enable Temurin build of x64/ppc64le/aarch64/s390x Linux for jdk-21+ using Adoptium DevKit
release:
- x64/ppc64le : gcc-11.3.0-Centos7.9.2009-b02
- aarch64 : gcc-11.3.0-Centos7.6.1810-b02
- s390x : s390x-on-s390x.RH7

aarch64 uses Centos7.6 as it is the matching latest version for RHEL7.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -12,6 +12,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+/* groovylint-disable MethodCount */
+
 import common.IndividualBuildConfig
 import common.MetaData
 import common.VersionInfo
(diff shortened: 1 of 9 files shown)


andrew-m-leonard:
run tests

karianna (approved):

sxa (commented):
We should be able to add in a `--with-devkit` parameter for s390x along with this too assuming some of my test runs overnight work as expected :-)

Also when https://github.com/adoptium/infrastructure/pull/3488 goes in and the build images get built the same will be possible on these three platforms too.

andrew-m-leonard:
> We should be able to add in a `--with-devkit` parameter for s390x along with this too assuming some of my test runs overnight work as expected :-)
> 
> Also when [adoptium/infrastructure#3488](https://github.com/adoptium/infrastructure/pull/3488) goes in and the build images get built the same will be possible on these three platforms too.

@sxa  That's great.
Since s390x won't be downloaded from devkit-runtimes, a PR will be needed to update --use-adoptium-devkit to search in /usr/local/devkit and verify tag, ditto for picking up the local x64/ppc64le/aarch64 ones..

Lets get this merged, and I can enable the later in a new PR.

sxa (approved):
LGTM. Note to reviewers - the s390x naming WILL be improved later for consistency but this will allow us to get parity for now.

PR #998, opened 2024-04-09
Correct missing openjdkbuild dir location for windows/mac signing

andrew-m-leonard:
With the change of openjdk build directory for jdk-21+ I missed one of the original path checks...!


steelhead31 (approved):
LGTM

sxa (approved):
We should get away with this since we're clearing out each time but I'll just note that if this is the directory that has the build target tuple bear in mind there can be more than one thing in there

PR #1003, opened 2024-04-11
Fix daily Slack status

andrew-m-leonard:
- Daily Slack status job fails to get the taggerdate of the latest build tag: https://ci.adoptium.net/view/Tooling/job/nightlyBuildAndTestStats_temurin/1466/console
taggerdate is only valid for annotated tags, upstream aarch32-jdk8u gets tagged with lightweight tags, to allow for both we should use creatordate
- non-tag driven release (Oracle managed releases) status not being published, eg.jdk22u
The logic for checking against the nonTagReleases had not being tested, been jdk22u the first to do this....


karianna (approved):

steelhead31 (approved):
LGTM

Issue #970, opened 2024-03-14
Additional test label for jdk20+


sophia-guo:
JDK20+ won't run on anything earlier than about RHEL7 .  Tests need to be limited to run on RHEL7 ( if it's hit a RHEL6/CentOS6 box that could be the problem  ).

@sxa

sxa:
SGTM - can you make the changes to add in the exclusion for machines with the appropriate labels in the pipelines?

Haroon-Khel:
> Tests need to be limited to run on RHEL7

Do you mean RHEL7+ or does this mean they cannot be run on RHEL/Centos 8?

sxa:
It will be fine on 8+. This is due to the glibc version on RHEL6 being too low for JDK21+ since [we switched that version for x64 to build on RHEL7](https://github.com/adoptium/ci-jenkins-pipelines/pull/627). The goal here is to stop it running on any Centos/RHEL6 system. They must still be able to run on other distributions (AFAIK all of our other ones in the CI are of a suitable glibc level - you can see which is in each on [our platforms wiki page](https://github.com/adoptium/infrastructure/wiki/End-of-support-date-for-OS-distributions))

PR #1005, opened 2024-04-11
Prevent jdk20+ linux jobs from running on rhel or centos 6 machines

Haroon-Khel:
ref https://github.com/adoptium/ci-jenkins-pipelines/issues/970


sxa (changes requested):
Can you adjust this so it's only on the Linux/x64 ones - we've never used any previous version of the OS for building on other architectures so this check is irrelevant.

sophia-guo (approved):

sxa (approved):

PR #1010, opened 2024-04-19
Update daily EA build Slack status to handle when GA builds are published

andrew-m-leonard:
- If the latest upstream tag is a GA build then don't expect an ea-beta build for it
- Also if a version has published a GA tag, then don't expect the ea-beta releases to be built for all the latest tags, as sometimes more than one build tag is published on a GA release
-


karianna (approved):

steelhead31 (approved):
lgtm

Issue #1007, opened 2024-04-12
automatic triggered builds "disabled tests" not early enough during release cycle, interferes with dry-run


andrew-m-leonard:
From 2024 April release dry run period, the current automatic disabling of tests from the "Saturday prior to release tuesday" is not early enough as it can interfere with dry-runs.
Suggest moving a week earlier, ie.2 Saturday's prior...

https://github.com/adoptium/ci-jenkins-pipelines/blob/799fc67d41035f1528c2d28c710d1622f27e90a4/pipelines/build/common/trigger_beta_build.groovy#L71

PR #1013, opened 2024-04-22
Update to compare against original scmReference with _adopt suffix

### LINKED ISSUE #1012: Expand comparison scmReference and JDK_BRANCH value from testenv.properties to cover Adoptium patch use case
The current code in openjdk_build.groovy that checks if we are using appropriately matched test material to the JDK code that we are building does not take into account the very occasional use case where we will float a patch in the Adoptium mirror repo ahead of it being upstreamed to the OpenJDK repository.  Since we are about to do just that for the April 2024 CPU release for s390x_linux patch into JDK 22, we need to update the comparison code.  

Related code:
https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/openjdk_pipeline.groovy#L80-L83
https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/build/openjdk_pipeline.groovy#L106-L114

smlambert:
Fixes #1012 

Note: This section of code will eventually be changed when https://github.com/adoptium/ci-jenkins-pipelines/issues/1007 is addressed.

### DIFF
--- a/pipelines/build/openjdk_pipeline.groovy
+++ b/pipelines/build/openjdk_pipeline.groovy
@@ -103,7 +103,7 @@ node('worker') {
             jdkBranch = resolveGaTag("${params.jdkVersion}", jdkBranch)
         }
 
-        if (jdkBranch == buildTag) {
+        if (jdkBranch == buildTag || jdkBranch == params.scmReference) {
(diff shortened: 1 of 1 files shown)


karianna (approved):

andrew-m-leonard (approved):
lgtm

PR #1015, opened 2024-04-22
Update textenv -ga tag check to handle jdk8u aarch32 port repo and checking for Lightweight git tags

andrew-m-leonard:
The jdk8u aarch32 repo is tagged with different different tag names, and uses git Lieghtweight tagging
Update the -ga commit search to handle the aarch32-port-jdk8u port repo, and the detection of Lightweight tags


steelhead31 (approved):
LGTM

sxa (approved):
Approving ... I'm assuming the correct number of backslashes has been verified for the grep operations here

PR #1019, opened 2024-04-24
Making jdk for jtreg build explicit

judovana:
and so imune against change of system jdk on underlying system/image


judovana:
This should fix the currently broken codetools - eg https://ci.adoptium.net/view/Dependencies/job/dependency_pipeline/1155/

It seesm that system jdk changed on that underlying system, and so old jtregs, were pulling to fresh jdk.

I'm a bit concerned about ` export PATH=$PATH:$JAVA_HOME/bin` becasue it is done in method, and that method is called several times, so at the end , there is `$PATH:$JAVA_HOME1/bin:$JAVA_HOME2/bin...:$JAVA_HOMEn/bin` .. which do not seem correct.

I will add a commits on this topic, but first let ssee if it fixes it.

karianna (approved):

sophia-guo (approved):

sophia-guo:
https://ci.adoptium.net/view/Dependencies/job/dependency_pipeline/1157/

judovana:
cool! ty!

Issue #1016, opened 2024-04-24
Update code-tools/jcstress to build and archive multiple jcstress.jar versions


smlambert:
Related to https://github.com/adoptium/aqa-tests/issues/5261, we would like to build several versions of jcstress.jar

Related code: https://github.com/adoptium/ci-jenkins-pipelines/tree/master/tools/code-tools

Recent run in Jenkins at https://ci.adoptium.net/view/Dependencies/job/dependency_pipeline/1155/artifact/jcstress/ shows what we currently build.

<img width="536" alt="Screenshot 2024-04-24 at 10 46 14 AM" src="https://github.com/adoptium/ci-jenkins-pipelines/assets/2836948/f6564bbe-6e17-48ce-b3a0-73535bc6272b">

PR #1020, opened 2024-04-25
Now building also jcstress-tests-all-20240222.jar 20220908.jar

judovana:
Those are based on closest (which were actually that day last) commit hashes. junit tests are passing.

In addition shortened hash to 7 latters and uses jdk17 to tip

Should be fixing https://github.com/adoptium/ci-jenkins-pipelines/issues/1016

### DIFF
--- a/tools/code-tools/jcstress.sh
+++ b/tools/code-tools/jcstress.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 
-###################################################################
-# Script to build jcstress to be reused by jdk testing community  #
-# currently builds tip and latest released version                #
-###################################################################
+#####################################################################
+# Script to build jcstress to be reused by jdk testing community    #
+# currently builds tip and latest released version                  #
+# It also build two snapshots, which are reused by adoptium testing #
+#####################################################################
 
 # shellcheck disable=SC2035,SC2155
 set -euo pipefail
@@ -21,7 +22,10 @@ function detectJdks() {
   find ${jvm_dir} -maxdepth 1 | sort
   echo "Available jdks 11 in ${jvm_dir}:"
   find ${jvm_dir} -maxdepth 1 | sort | grep -e java-11- -e jdk-11
+  echo "Available jdks 17 in ${jvm_dir}:"
+  find ${jvm_dir} -maxdepth 1 | sort | grep -e java-17- -e jdk-17
   jdk11=$(readlink -f $(find ${jvm_dir} -maxdepth 1 | sort | grep -e java-11- -e jdk-11 | head -n 1))
+  jdk17=$(readlink -f $(find ${jvm_dir} -maxdepth 1 | sort | grep -e java-17- -e jdk-17 | head -n 1))
 }
 
 REPO_DIR="jcstress"
@@ -36,7 +40,7 @@ detectJdks
 pushd $REPO_DIR
   git checkout master
   tip=`git log | head -n 1 | sed "s/.*\s\+//"` || true
-  tip_shortened=`echo ${tip:0:10}`
+  tip_shortened=`echo ${tip:0:7}`
   latestRelease=`git tag -l | sort -Vr | head -n 1`
   rc=$main_name-$latestRelease.jar
 
@@ -51,11 +55,25 @@ pushd $REPO_DIR
 
   # tip
   git checkout master
-  export JAVA_HOME=$jdk11
+  export JAVA_HOME=$jdk17
   mvn clean install
   mv tests-all/target/$main_file $main_name-$tip_shortened.jar
   echo "Manually renaming $main_name-$tip_shortened.jar as $main_name-tip.jar to provide latest-unstable-recommended file"
   ln -fv $main_name-$tip_shortened.jar $main_name-tip.jar
+  mvn clean
+
+  # 20240222
+  git checkout c565311051494f4b9f78ec86eac6282f1de977e2
+  export JAVA_HOME=$jdk11
+  mvn clean install
+  mv -v tests-all/target/$main_name.jar jcstress-tests-all-20240222.jar
+  mvn clean
+
+  # 20220908
+  git checkout d118775943666d46ca48a50f21b4e07b9ec1f7ed
+  export JAVA_HOME=$jdk11
+  mvn clean install
+  mv -v tests-all/target/$main_name.jar jcstress-tests-all-20220908.jar
   mvn clean
 
   echo "Resetting repo back to master"
--- a/tools/code-tools/jcstress.sh
+++ b/tools/code-tools/jcstress.sh
@@ -28,6 +28,20 @@ function detectJdks() {
   jdk17=$(readlink -f $(find ${jvm_dir} -maxdepth 1 | sort | grep -e java-17- -e jdk-17 | head -n 1))
 }
 
+function buildJcstress {
+  local jdk="${1}"
+  local checkout="${2}"
+  local target="${3}"
+  export JAVA_HOME="${jdk}"
+  mvn clean
+  git checkout "${checkout}"
+  mvn clean install
+  mv -v "tests-all/target/${main_name}.jar" "${target}"
+  mvn clean
+  git checkout master
+  unset JAVA_HOME
+}
+
 REPO_DIR="jcstress"
 main_name=$REPO_DIR
 main_file=$main_name.jar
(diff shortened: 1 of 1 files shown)


tools/code-tools/jcstress.sh
```diff
@@ -36,29 +54,25 @@ detectJdks
...
-  mvn clean install
-  mv tests-all/target/$main_file $main_name-$tip_shortened.jar
+  buildJcstress "${jdk17}" "master" "${main_name}-${tip_shortened}.jar"
   echo "Manually renaming $main_name-$tip_shortened.jar as $main_name-tip.jar to provide latest-unstable-recommended file"
   ln -fv $main_name-$tip_shortened.jar $main_name-tip.jar
-  mvn clean
 
-  echo "Resetting repo back to master"
-  git checkout master
+  # 20240222
```
karianna, line 71: What is the significance of that date?  Is that a tag for a major version?

judovana: As I see it, there is no significance. Unluckily there is also no tag. Doule unluckily jcstress tags extremely rarely:(

As described here: https://github.com/adoptium/aqa-tests/issues/5261 ;  the jcstres in aqavit have heavily unclear binaries to execute:

> The getDepenencies is pulling jcstress-tests-all-20240222.jar'
https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L182
in addition, it is pulling it from personal server

and

> however, the https://github.com/adoptium/aqa-tests/blob/master/system/jcstress/playlist.xml
is using jcstress-tests-all-20220908.jar

Those are not base don any tag, nor hash.  They will be reused in the cstress runner, so the  https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L182 and  https://github.com/adoptium/aqa-tests/blob/master/system/jcstress/playlist.xml are in harmony.

I think first step will be to let the getDependencies.pl to download 20220908, and use that in playlist.xml. Then to bump them both to 20240222 (as this is quite fresh). There may come som if jdk 8/11/...22 swithc as is done with jtrregs. This is stil to be decided. For now, I wanted to prepare binareis to use.

Thanx for bringing it up. If you have any top level hints, dont hesitate to write them to https://github.com/adoptium/ci-jenkins-pipelines/issues/1016 and  https://github.com/adoptium/aqa-tests/issues/5261.
Actually.. any hints welcomed :) Thanx for eyball!

smlambert:
We do not need a build from that date.  We need/want to use tagged or released builds.  In order to do that, we need to 'test' that an appropriately recent tagged build can run the test targets we have, without major errors.

So indeed, please do not bother building untagged/untracked versions of jcstress.  I do not have time to review this PR today as I am on 'personal time off'.  Please do not rush into merging anything.  There is all the time in the world to get things right and yet takes no time at all to do the wrong thing.

judovana:
This PR is not in rush for sure. Whatever we agree should be built, will be built.

However I think tagged builds are not an option. jcstress team (read shade) tags somehow randomly, and last tag is pretty old.
Also every time I reported an issue, first order was update jcstress, and if not to some particular commit,then to head. And indeed it worked in 3/4 of cases. 
Similarly the https://mvnrepository.com/artifact/org.openjdk.jcstress is doing that. IIRC, I was trying to do maven releases more often, and I still have permissions to do that, but that is still jsut random local build.  Same counts for JMH. And you still one of  project oweners to tag the repo, and that may be show-stopper pretty often. 

Also I doubt we wish to go with tip, as that can indeed change dramatically without warning. From that point of view the  20240222 is pretty good option as it is known to be stable and is really fresh.

I was following https://github.com/adoptium/aqa-tests/issues/5261  where you gave me green light, unluckily only during voice chat, and it literally says:  

> improve https://ci.adoptium.net/view/Dependencies/job/dependency_pipeline/lastSuccessfulBuild/artifact/jcstress/ so it builds jcstress-tests-all-20220908.jar and jcstress-tests-all-20240222.jar and tip

The itp as is now is already  building tip.

Which is what this PR is doing (tip was already there).  

If you reconsidered, it is ok and I will follow for sure, I do not have intentions to break anything, but I would like to have aqa-tests jcstress build on better foundation then on random  jcstress-tests-all-20220908.jar which was downloaded ages ago and can no longer be recreated. (Or I have not found it).
So indeed the builder should be  building 20220908 - so the https://github.com/adoptium/aqa-tests/blob/master/system/jcstress/playlist.xml  and  https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L182 can immediately reuse that. Then it should be able to pick up  20240222 moreover out of the box. Maybe except jdk8 and maybe only on demand.. that will follow.

Generally it is good we had this conversation. I realised, that all above is valid also for JMH, and that  jcstres palylist and/or https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.xml have to be able to work - on demand only! - with jcstress tip, as that is waht jcstress people will always ask  us to do first.

No rush from me. Do not spoil your time off. You made yourself clear you wish to double check that, and I'm greatful for it.

judovana:
Looking to https://github.com/adoptium/aqa-tests/issues/5278 (and seeing https://github.com/adoptium/aqa-tests/pull/5270 work ok) This shold be building only  20240222, and lets say with "correct" name of  `jcstress-tests-all-20240222.jar` ?

sophia-guo:
Currently aqa-tests is grabing jcstress from personal server so this PR won't affect aqa-tests. I think 20240222 is enough ( no need to build 20220908.jar). 

What's the difference of building the jar with jdk17 and jdk11? Do we need both version? Once this is in we can update aqa-tests.

judovana:
> Currently aqa-tests is grabing jcstress from personal server so this PR won't affect aqa-tests. I think 20240222 is enough ( no need to build 20220908.jar).

Right and right. I will drop the 2022 version as the 2024 is already "iun production" (was not when I strarted thi sPR)
But immediately once it is in I will switch the wget of  it as descibed in https://github.com/adoptium/aqa-tests/issues/5261 from personal server, to result of this build (latestSucesfullBuild jcstress/Artifacts)

> 
> What's the difference of building the jar with jdk17 and jdk11? Do we need both version? Once this is in we can update aqa-tests.

That appeared to be wrong assumption. I will swithc it all to jdk8.At least we will know, when jdk8  is not worthy anymore.

judovana:
Please double check the naming of main tarball.
I'm using: jcstress.jar,  jcstress-0.16.jar jcstress-20240222.jar, jcstress-somehash.jar and jcstress-tip.jar, (last two and first two are hardlink). The someHash one is now `jcstress-d6a81c3.jar`

Where this is more aligned to how maven is generating the jars, current  usage in adoptium is `jcstress-tests-all-20240222.jar`  which is slightly more describing waht *directroy* the tests are from. However I think it is not necessary  to highlight because it is the only jar produced during default build. Note, that the old naming is used in both current generators (see https://github.com/adoptium/aqa-tests/issues/5278). But that woudl be super easy to adjust onc ethis is in. TY!

sophia-guo:
> That appeared to be wrong assumption. I will swithc it all to jdk8.At least we will know, when jdk8 is not worthy anymore.

Not sure what do you mean switch it all to jdk8. The project mentioned `The project requires JDK 11+ to build.` https://github.com/openjdk/jcstress ?

sophia-guo:
https://ci.adoptium.net/view/Dependencies/job/dependency_pipeline/1186/artifact/jcstress/

Are jcstress-0.16.jar, jcstress-20240222.jar, jcstress-d6a81c3.jar  different? If not we don't need to keep all of them?

Could you @judovana make the PR ready for review if it's ready?

judovana:
The linking follows all other ci-pipellines buildjobs.
 * tip have hash in name, and is linked as tip (jcstress-d6a81c3.jar +jcstress-tip.jar)
 * latest release have release in name and is linked as "that"  (jcstress-0.16.jar + jcstress.jar)
 * jcstress-20240222.jar as artificial  jar for internal purposes  have no linking

The goal is to have predictable names for pernament links, and still haveing clear what is waht.

The  jcstress-20240222.jar is now very similar to jcstress-d6a81c3.jar because only few commits was added in meantime, but it is ill idea to have internal testing agasint tip.

sophia-guo:
So jcstress.jar, jcstress-d6a81c3.jar and jcstress-tip.jar are same? For now I think tip, latest release, jcstress-20240222.jar are enough as long as the sha is provided.

tools/code-tools/jcstress.sh
```diff
@@ -19,11 +20,34 @@ function hashArtifacts() {
...
+  echo "Available jdks 8 in ${jvm_dir}:"
+  echo "  Note, that you need at at least jdk 11 to build 20240222, tip and 0.16. Listing all jdks anyway."
+  find ${jvm_dir} -maxdepth 1 | sort | grep -e java-1.8.0- -e jdk-1.8.0 -e java-8- -e jdk-8
   echo "Available jdks 11 in ${jvm_dir}:"
   find ${jvm_dir} -maxdepth 1 | sort | grep -e java-11- -e jdk-11
+  echo "Available jdks 17 in ${jvm_dir}:"
+  find ${jvm_dir} -maxdepth 1 | sort | grep -e java-17- -e jdk-17
+  jdk8=$(readlink -f $(find ${jvm_dir} -maxdepth 1 | sort | grep -e java-1.8.0- -e jdk-1.8.0 -e java-8- -e jdk-8 | head -n 1))
   jdk11=$(readlink -f $(find ${jvm_dir} -maxdepth 1 | sort | grep -e java-11- -e jdk-11 | head -n 1))
+  jdk17=$(readlink -f $(find ${jvm_dir} -maxdepth 1 | sort | grep -e java-17- -e jdk-17 | head -n 1))
```
sophia-guo, line 32: Do we still need this?

sophia-guo (approved):

judovana:
> So jcstress.jar, jcstress-d6a81c3.jar and jcstress-tip.jar are same? For now I think tip, latest release, jcstress-20240222.jar are enough as long as the sha is provided.

no  jcstress-d6a81c3.jar and jcstress-tip.jar  are same, and then jcstress.jar anf (jcstress-0.16.jar as I wrote above.

All other ci-tools are keeping this convention. So I would keep it.

judovana:
tyvm! for approve!

sophia-guo:
@karianna any suggestion?

tools/code-tools/jcstress.sh
karianna: Are we missing Java 21 here as well?

tools/code-tools/jcstress.sh
karianna: @smlambert - I think I need your guidance here :-).  Which version(s) of jcstress does the testing group need?  I read the other issues but I'm still unsure...

karianna (commented):
Need @smlambert to confirm versions we want supported.

tools/code-tools/jcstress.sh
judovana: What do you mean miss java 21?  The container have jdk8, jdk11, jdk17 and also jdk21 iirc. Jcsress since some 2023 needs at elast jdk11 to build with. I'm checking the hdk versiosn just for case the build goes wrong. I can check the 21 too if you wish.

the upstream is correctlysetting source/version/release javc flags so to build each version by different jdk was just candy on the top.   Now all three jcstresses (latest release, tip, and  "latest used by us" )  are built by  jdk11. If you wish, I have no objections to build tip by soemthign newer. :) ty!

tools/code-tools/jcstress.sh
judovana: as any builder, we need tip and latest release for sure.   I heavily recommend to not get stick to latest release, as they are not fluently released, and to verify jcstress failure you have to try something newer, always, even tip from time to time.   So A middle path is to pick up something known to be stable on the fly, and starting using that. 20240202 is pretty fresh, and known to be stable, so very good candidate.

tools/code-tools/jcstress.sh
karianna: It sounds like you only need to check for the existence of 11+ in that case?  It seems over kill to check for each LTS version...

smlambert:
Yes, in absence of a stable release tag to build off of, we proceed with this approach.  If jcstress begins to employ release tagging in future, we can adjust.

PR #1021, opened 2024-04-30
Added JMH to be also built by code-tools

judovana:
JMH is another code-tools openjdk project which release pretty randomly.
JMH suite itslef is engine to run benchamrks,  and reelased versions are apearing in maven repos (similarly as jcstress).
JMH samples can be used as pretty good testsuite when reduced to run minimal iterations.

Generally spoken the build of TIP is heavily missed world wide.

### DIFF
--- a/tools/code-tools/Jenkinsfile
+++ b/tools/code-tools/Jenkinsfile
@@ -61,6 +61,14 @@ pipeline {
                         build('jcov')
                     }
                 }
+                stage('jmh') {
+                    agent {
+                        label NODE_LABEL
+                    }
+                    steps {
+                        build('jmh')
+                    }
+                }
                 stage('temurin-sbom') {
                     agent {
                         label NODE_LABEL
--- a/tools/code-tools/jmh.sh
+++ b/tools/code-tools/jmh.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+################################################################
+# Script to build jmh to be reused by jdk testing community    #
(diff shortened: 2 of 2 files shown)


karianna:
Is this actually required.  Aleksey regularly releases this upstream

judovana:
> Is this actually required. Aleksey regularly releases this upstream

I wish it is not requored:(

https://github.com/openjdk/jmh/tags - year ago 
followed by https://github.com/openjdk/jmh/tags

But true - latest released is brobably not necessary. only tip.

judovana:
Actually one reason to build both master and latest released - if the build of tip fails, it may behard to say it it is jdk or jmh who isguilty. If fails latest released, then it is most likely jdk.

gdams:
ping @shipilev to see whether he does plan to continue building this or not

sophia-guo:
Is this still in progress as jenkinsfile also need to update if we'd like to build here?

judovana:
> Is this still in progress as jenkinsfile also need to update if we'd like to build here?

Nope. sorry. I forget about it.

shipilev:
The overwhemling majority of users consume JMH through Maven Central releases. So, unless you are planning to publish SNAPSHOT versions somewhere, I see no particular value in building JMH separately in Adoptium. I see no particular cost for it either, so it is your choice, really.

sophia-guo (approved):

judovana:
tyvm!

smlambert (approved):

PR #1024, opened 2024-05-02
add riscv64 to JDK17 release pipeline

sxa:
This will be formally discussed at the next PMC meeting before performing a release, but this change will allow us to be able to release a JDK17 on Linux/riscv64 which is believed to be in a good state now, and this will allow us to fully prep and test a release prior to shipping.

### DIFF
--- a/pipelines/jobs/configurations/jdk17u_release.groovy
+++ b/pipelines/jobs/configurations/jdk17u_release.groovy
@@ -35,7 +35,6 @@ targetConfigurations = [
         'riscv64Linux': [
                 'temurin'
         ]
-
 ]
 
 return this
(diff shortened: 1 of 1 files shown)


pipelines/jobs/configurations/jdk17u_release.groovy
```diff
@@ -31,7 +31,11 @@ targetConfigurations = [
         ],
         'arm32Linux'  : [
                 'temurin'
+        ],
+        'riscv64Linux': [
+                'temurin'
         ]
+
```
karianna, line 38:
```suggestion
```

karianna (approved):

gdams (approved):
LGTM (once you've addressed Martijn's change)

PR #1025, opened 2024-05-03
don't pass node label to windows installer job

gdams:
Now that we define the installer job as a [Jenkinsfile](https://github.com/adoptium/installer/blob/master/wix/Jenkinsfile) it makes more sense to define this there.

Also fixes JVM which we've been incorrectly passing in as `temurin` for some time now


tellison (approved):

steelhead31 (approved):
LGTM

PR #1043, opened 2024-06-13
Update DevKit build "bootstrap" to ensure all tooling comes from bootstrap

andrew-m-leonard:
All the necessary DevKit tooling (ld, as, ar, ranlib,...) needs explicitly setting to the bootstrap to ensure the final DevKit is bootstrapped correctly.

### DIFF
--- a/pipelines/build/devkit/make_devkit.sh
+++ b/pipelines/build/devkit/make_devkit.sh
@@ -70,7 +70,7 @@ mv build/devkit/result/${devkit_target}-to-${devkit_target} ${BOOTSTRAP_DEVKIT}
 
 # Make final "DevKit" using the bootstrap devkit
 rm -rf build/devkit
-echo "Building 'final' DevKit toolchain, using 'bootstrap' toolchain in $(pwd)/build/bootstrap_${devkit_target}-to-${devkit_target}"
+echo "Building 'final' DevKit toolchain, using 'bootstrap' toolchain in ${BOOTSTRAP_DEVKIT}"
 cd make/devkit && pwd && \
   LD_LIBRARY_PATH="${BOOTSTRAP_DEVKIT}/lib64:${BOOTSTRAP_DEVKIT}/lib" \
   PATH="${BOOTSTRAP_DEVKIT}/bin:$PATH" \
@@ -79,7 +79,7 @@ cd make/devkit && pwd && \
        CXX=${BOOTSTRAP_DEVKIT}/bin/g++ \
        LD=${BOOTSTRAP_DEVKIT}/bin/ld \
        AR=${BOOTSTRAP_DEVKIT}/bin/ar \
-       AS=${BOOTSTRAP_DEVKIT}/bin/AS \
+       AS=${BOOTSTRAP_DEVKIT}/bin/as \
        RANLIB=${BOOTSTRAP_DEVKIT}/bin/ranlib \
        OBJDUMP=${BOOTSTRAP_DEVKIT}/bin/objdump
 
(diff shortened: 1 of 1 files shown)


karianna (approved):

andrew-m-leonard:
@sxa for review please

sxa (commented):
@andrew-m-leonard Is this inteded to resolve the glibc dependency issue while building the devkit? I tried it on F38 and it seems to hit the same issue.

sxa (commented):
I think this is ok but to be clear what was the reason for changing from the environment variables for CC/CXX to putting them on the `make` command line? Was that significant? Is there any potential problem if, say, the compiler calls `ld` internally that might result in it not using the one from the devkit in this situation?
It may be nothing, but just a though when I saw the changing in where it is defined.

andrew-m-leonard:
> I think this is ok but to be clear what was the reason for changing from the environment variables for CC/CXX to putting them on the `make` command line? Was that significant? Is there any potential problem if, say, the compiler calls `ld` internally that might result in it not using the one from the devkit in this situation? It may be nothing, but just a though when I saw the changing in where it is defined.

@sxa  Which do you think is better, export or make variable ? or should we just do both...?

sxa:
> > I think this is ok but to be clear what was the reason for changing from the environment variables for CC/CXX to putting them on the `make` command line? Was that significant? Is there any potential problem if, say, the compiler calls `ld` internally that might result in it not using the one from the devkit in this situation? It may be nothing, but just a though when I saw the changing in where it is defined.
> 
> @sxa Which do you think is better, export or make variable ? or should we just do both...?

Honestly I'm pretty confident that once you've set the PATH you're fine if the tools aren't suffixed (i.e. they don't have `ld-2.39` instead of just `ld`). I've generally used the environment variables rather than parameters to make (which is why I asked the question originally rather than me having deep insight here!) but that's what you had originally (and I guess you saw some issues with it?) From a quick test it seems to manage to find them in the PATH even if `CC` etc. are set to gibberish values ...

andrew-m-leonard:
> > > I think this is ok but to be clear what was the reason for changing from the environment variables for CC/CXX to putting them on the `make` command line? Was that significant? Is there any potential problem if, say, the compiler calls `ld` internally that might result in it not using the one from the devkit in this situation? It may be nothing, but just a though when I saw the changing in where it is defined.
> > 
> > 
> > @sxa Which do you think is better, export or make variable ? or should we just do both...?
> 
> Honestly I'm pretty confident that once you've set the PATH you're fine if the tools aren't suffixed (i.e. they don't have `ld-2.39` instead of just `ld`). I've generally used the environment variables rather than parameters to make (which is why I asked the question originally rather than me having deep insight here!) but that's what you had originally (and I guess you saw some issues with it?) From a quick test it seems to manage to find them in the PATH even if `CC` etc. are set to gibberish values ...

The issues we had were from a 3rd party (in this case Red Hat) independent DevKit build, which was different, and although we've tracked down the exact difference, I want to try and do "belt & braces" on all the variables, i'm thinking maybe changing to PATH and export... what do you think?

sxa (approved):
> The issues we had were from a 3rd party (in this case Red Hat) independent DevKit build, which was different, and although we've tracked down the exact difference, I want to try and do "belt & braces" on all the variables, i'm thinking maybe changing to PATH and export... what do you think?

If the "exact difference" referenced above was unrelated to these variables I'd personally prefer not to set them unless there was a known issue with the previous `PATH` setting not doing what you expect. I'd be nervous that adding in potentially unnecessary definitions into this (and increasing code complexity) may lead people to a false sense that changing them in the future might do "something".

Having said that, I don't want to block on this, so I can approve regardless as I don't think it'll actively damage anything.

PR #1046, opened 2024-06-18
Update trigger_beta_build and daily Slack status jobs to handle jdk(head) version branching

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3856

- Correct EA trigger jenkins job to correctly find and build jdk(head) version tags
- Correct daily Slack status to report correct jdk(head) version build status
-

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -622,7 +622,7 @@ node('worker') {
         }
 
         // Slack message:
-        slackSend(channel: slackChannel, color: statusColor, message: 'Adoptium last 7 days Overall Build Success Rating : *' + variant + '* => *' + overallNightlySuccessRating + '* %\n  Build Job Rating: ' + totalBuildJobs + ' jobs (' + nightlyBuildSuccessRating.intValue() + '%)  Test Job Rating: ' + totalTestJobs + ' jobs (' + nightlyTestSuccessRating.intValue() + '%) <' + BUILD_URL + '/console|Detail>')
+        //slackSend(channel: slackChannel, color: statusColor, message: 'Adoptium last 7 days Overall Build Success Rating : *' + variant + '* => *' + overallNightlySuccessRating + '* %\n  Build Job Rating: ' + totalBuildJobs + ' jobs (' + nightlyBuildSuccessRating.intValue() + '%)  Test Job Rating: ' + totalTestJobs + ' jobs (' + nightlyTestSuccessRating.intValue() + '%) <' + BUILD_URL + '/console|Detail>')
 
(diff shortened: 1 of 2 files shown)


steelhead31 (approved):
Looks ok :)

andrew-m-leonard:
@sxa i've upgraded super-linter to 6.6.0, as the current version 5.0.0 was "hanging" in the groovy-linter.
Seems fine with 6.6.0

tools/nightly_build_and_test_stats.groovy
```diff
@@ -56,35 +51,50 @@ def isGaTag(String version, String tag) {
...
 
-    // Need to include jdk8u to avoid picking up old tag format    
-    def jdk8Filter = (version.contains("jdk8u")) ? "| grep 'jdk8u'" : ""
+    return openjdkRepo
+}
+
+// Get the latest upstream openjdk build tag
+def getLatestOpenjdkBuildTag(String version) {
+    def openjdkRepo = getUpstreamRepo(version)
+    def versionInt = (version == "aarch32-jdk8u" || version == "alpine-jdk8u") ? 8 : version.replaceAll("u", "").replaceAll("jdk", "").toInteger()
```
sxa, line 78: Could these two `replaceAll` invocations  be replaced with the simpler `replaceAll("[a-z]","");` to strip out all letters from the name?

tools/nightly_build_and_test_stats.groovy
```diff
@@ -56,35 +51,50 @@ def isGaTag(String version, String tag) {
...
+
+    def tagFilter
     if (version == "aarch32-jdk8u") {
-        jdk8Filter += " | grep '\\-aarch32\\-'"
+        tagFilter = "| grep 'jdk8u.*-aarch32-'"
+    } else if (version.contains("jdk8u")) {
+        tagFilter = "| grep 'jdk8u'"
+    } else {
+        tagFilter = "| grep 'jdk-"+versionInt+"[\\.\\+].*'"
```
sxa, line 86: If I'm reading this right you're looking explicitly to match e.g. `jdk-11.` or `jdk-11+` at the start, so the `.*` at the end would seem to be superfluous.
```suggestion
        tagFilter = "| grep 'jdk-"+versionInt+"[\\.\\+]'"
```

sxa (approved):
Couple of suggestions which I'd prefer to see addressed if possible, but since it's close to the end of my day I don't want to block this from being merged when there is the potential for a relevant tag to come out, so I'll give it my approval.

tools/nightly_build_and_test_stats.groovy
andrew-m-leonard: I think so

tools/nightly_build_and_test_stats.groovy
andrew-m-leonard: absolutely, thanks

andrew-m-leonard:
> Noting that there is probably some doc that needs updating based on all the changes from this new process.
yep, already updated: https://github.com/adoptium/temurin-build/wiki/Setting-up-new-build-jobs,-mirrors,...-when-a-new-openjdk-version-is-branched-from-jdk(head)

PR #1052, opened 2024-06-21
Save DevKit downloaded rpms from bootstrap build to save download in final build

andrew-m-leonard:
Update make_devkit.sh to use the downloaded rpms from the bootstrap gcc build, to use in the 2nd final build

Note: Disabled new super-linter 6.6.0 VALIDATE_SHELL_SHFMT, as not useful

### DIFF
--- a/pipelines/build/devkit/make_devkit.sh
+++ b/pipelines/build/devkit/make_devkit.sh
@@ -43,16 +43,16 @@ patch -p1 < ../Tools.gmk.patch
 
 devkit_target="${ARCH}-linux-gnu"
 
-if [ "${BASE_OS}" = "rhel" ]; then
-  mkdir -p ../../../build/devkit/${VERSION}/build/devkit/download/rpms/s390x-linux-gnu-Centos${BASE_OS_VERSION}
-  # Downlod RPMS from RHEL (Requires machine to be attached to RHEL subscription)
+if [ "${BASE_OS}" = "rhel" ] && [ "${ARCH}" = "s390x" ]; then
+  mkdir -p build/devkit/download/rpms/s390x-linux-gnu-Centos${BASE_OS_VERSION}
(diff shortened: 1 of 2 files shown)


pipelines/build/devkit/make_devkit.sh
```diff
@@ -66,10 +66,18 @@ cd make/devkit && pwd && make TARGETS=${devkit_target} BASE_OS=${BASE_OS} BASE_O
...
+
+mv build/devkit/result/${devkit_target}-to-${devkit_target} "${BOOTSTRAP_DEVKIT}"
+mv build/devkit/download/rpms/${ARCH}-linux-gnu-Centos${BASE_OS_VERSION} "${BOOTSTRAP_DOWNLOADED_RPMS}"
 
 # Make final "DevKit" using the bootstrap devkit
 rm -rf build/devkit
+
+# Move saved bootstrap rpm downloads to final build folder
+mkdir -p build/devkit/download/rpms
+mv ${BOOTSTRAP_DOWNLOADED_RPMS} build/devkit/download/rpms/${ARCH}-linux-gnu-Centos${BASE_OS_VERSION}
```
sxa, line 79: My gut feel is that maybe using `ln` to make them appear in both locations might be nicer, but take or leave this as you wish.
```suggestion
ln ${BOOTSTRAP_DOWNLOADED_RPMS} build/devkit/download/rpms/${ARCH}-linux-gnu-Centos${BASE_OS_VERSION}
```

sxa (approved):
LGTM, although I'd prefer that the linter fix is split out (Particularly since this is affecting a load of other PRs so having a commit in the repo explicitly to fix that would be useful to be able to reference, which we could also post so Slack for awareness from others :-) )

karianna (approved):

PR #1055, opened 2024-06-24
avoids image name error when pulling from non-default registery

mahdipub:
when pulling from a non-default registery, you will have local image in full name include docker registery in it. But the whole pipeline works with DEFAULT behaviour when pulling from dockerhub that creates local image with image name only. This change, adds a tag to pulled local image with image_name only and avoids issue in continue. Also as a result of changes we do not need to clean `dockerImageDigest` later in the code. So that line removed.

Signed-off-by: mahdi@ibm.com

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -2044,7 +2044,9 @@ class Build {
                                 }
                             }
                             // Store the pulled docker image digest as 'buildinfo'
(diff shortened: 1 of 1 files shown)


karianna (approved):

mahdipub:
As not all docker images has `.RepoDigests` then counting on index 0 of it could cause issues like:
```
docker inspect '--format={{index .RepoDigests 0}}' adoptopenjdk/centos7_build_image_cuda

template parsing error: template: :1:2: executing "" at <index .RepoDigests 0>: error calling index: reflect: slice index out of range
```
So I am going to close this and reopen another one without that code improvement change.

mahdipub:
other PR without code improvement https://github.com/adoptium/ci-jenkins-pipelines/pull/1076

PR #1069, opened 2024-07-03
Use JDKxu config over JDKx config wherever possible

adamfarley:
The affected code seems to use JDKx config even if the more recent jdkxu config is available.

This is incorrect, and should be resolved.

Actioned to resolve the General retrospective comment [here](https://github.com/adoptium/temurin/issues/28#issuecomment-2042865717).

### DIFF
--- a/pipelines/build/openjdk_pipeline.groovy
+++ b/pipelines/build/openjdk_pipeline.groovy
@@ -190,14 +190,14 @@ node('worker') {
     }
 
     // Load buildConfigFilePath. This is where jdkxx_pipeline_config.groovy is located. It contains the build configurations for each platform, architecture and variant.
-    def buildConfigFilePath = (params.buildConfigFilePath) ?: "${DEFAULTS_JSON['configDirectories']['build']}/${javaToBuild}_pipeline_config.groovy"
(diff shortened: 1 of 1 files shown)


adamfarley:
run tests

pipelines/build/openjdk_pipeline.groovy
```diff
@@ -190,14 +190,14 @@ node('worker') {
     }
 
     // Load buildConfigFilePath. This is where jdkxx_pipeline_config.groovy is located. It contains the build configurations for each platform, architecture and variant.
-    def buildConfigFilePath = (params.buildConfigFilePath) ?: "${DEFAULTS_JSON['configDirectories']['build']}/${javaToBuild}_pipeline_config.groovy"
+    def buildConfigFilePath = (params.buildConfigFilePath) ?: "${DEFAULTS_JSON['configDirectories']['build']}/${javaToBuild}u_pipeline_config.groovy"
```
karianna, line 193: If the u_pipeline_config isn't there, what does buildConfigFilePath fall back to?

adamfarley: If it isn't there, then the if statement in the following lines of code should notice that and default to the non-u pipeline config file.

sxa (approved):

adamfarley:
@karianna - Does this address your concern? Are we good to merge?

karianna (approved):

Issue #1068, opened 2023-12-15
Colour of smoke test block in build pipeline does not reflect failed status correctly.


sxa:
**What are you trying to do?** Run a build and see the status reflected in the blocks in the pipeline overview

**Expected behaviour:** If the smoke test failes causing an overall fail status in the pipeline, then the corresponding square in the pipeline should show  failed status

**Observed behaviour:**
![image](https://github.com/adoptium/temurin-build/assets/6487691/6c9e1fdf-9e1f-47be-adde-d8a289b509d3)
This screenshot is despite the failure of [riscv64-temurin_SmokeTests#14](https://ci.adoptium.net/job/build-scripts/job/jobs/job/evaluation/job/jobs/job/jdk/job/jdk-evaluation-linux-riscv64-temurin_SmokeTests/14) triggered by [jdk-evaluation-linux-riscv64-temurin#47](https://ci.adoptium.net/job/build-scripts/job/jobs/job/evaluation/job/jobs/job/jdk/job/jdk-evaluation-linux-riscv64-temurin/47/console) (it's the overview for jdk-evaluation-linux-riscv64-temurin where the screen shot above was visible)

**Any other comments:**

sxa:
Related:
- https://github.com/adoptium/ci-jenkins-pipelines/issues/234
- https://github.com/adoptium/ci-jenkins-pipelines/issues/11

andrew-m-leonard:
This issue has been frustrating me too! i'm working on a fix

PR #1071, opened 2024-07-04
Fix up pipeline stage result setting and exception handling

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/1068

Problem:
- Because for several sub-jobs like "tests", we use propagate: false, the job result does not set the stage result, we currently only set the overall build result. (Note: We want propagate: false so we can run all the tests and then report all the failures, rather than exit the job at the very first failure)

This PR fixes:
- Ensure all sub-jobs called within "stages" also set the stage result
- Tidy up several catch(Exception) clauses that swallow exceptions, and were not setting overall build result to FAILURE
- Add missing "sign verification" stage

**Fixed tests:**
- Various x64Linux forced failures/aborts, showing correct SmokeTest stage results:
Screenshot 2024-07-04 at 13 06 55
- Various Mac stages, aborted/failed, all showing correct Stage results:
Screenshot 2024-07-04 at 12 39 21

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -318,7 +318,7 @@ class Build {
     def setStageResult(String stage, String result) {
         if (result != "SUCCESS") {
             // Use catchError to issue error message and set build & stage result
-            catchError(buildResult: result, stageResult: result) {
+            context.catchError(buildResult: result, stageResult: result) {
                 context.error("${stage} not successful, setting stage result to: "+result)
             }
         } else {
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

sxa (approved):
I'm not an expert on these pipelines, but if this definitely works I'm delighted to approve it (And it'll be fairly obvious soon if there are any problems with it I expect)

sxa:
Adding Sophia as a potential reviewer as she was involved in the issue referenced in the description.

sophia-guo (approved):
LGTM

PR #1076, opened 2024-07-09
avoids image name error when pulling from non-default registery

mahdipub:
when pulling from a non-default registery, you will have local image in full name include docker registery in it. But the whole pipeline works with DEFAULT behaviour when pulling from dockerhub that creates local image with image name only. This change, adds a tag to pulled local image with image_name only and avoids issue in continue.

Signed-off-by: mahdi@ibm.com

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -2044,6 +2044,8 @@ class Build {
                                 }
                             }
                             // Store the pulled docker image digest as 'buildinfo'
+                            def long_docker_image_name = context.sh(script: "docker image ls | grep ${buildConfig.DOCKER_IMAGE} | head -n1 | awk '{print \$1}'", returnStdout:true).trim()
+                            context.sh(script: "docker tag '${long_docker_image_name}' '${buildConfig.DOCKER_IMAGE}'", returnStdout:false)
                             dockerImageDigest = context.sh(script: "docker inspect --format='{{.RepoDigests}}' ${buildConfig.DOCKER_IMAGE}", returnStdout:true)
 
                             // Use our dockerfile if DOCKER_FILE is defined


karianna (approved):

sxa:
Hi @mahdipub I'm struggling a bit to understand the description of this PR. Can you provide an example of what situation this is fixing? I just want to make sure this is clear before approving. We have some cider in first handling local images and I want to make sure this doesn't result in any conflicts :-)

mahdipub:
> Hi @mahdipub I'm struggling a bit to understand the description of this PR. Can you provide an example of what situation this is fixing? I just want to make sure this is clear before approving. We have some cider in first handling local images and I want to make sure this doesn't result in any conflicts :-)

@sxa Sure. Here is an example conf we set for an image which comes from local artifactory (not dockerhub):
https://github.com/ibmruntimes/ci-jenkins-pipelines/blob/7098566e754a889b797b883e512303564cae1739/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy#L124

```
### jdk11u_pipeline_config.groovy

     dockerImage: 'sys-rt-docker-local/semeru/s390_rhel7_build_image',
     dockerRegistry: 'https://docker-na.artifactory.swg-devops.com/',
```

When we pull image, if the registry is dockerhub, then local image name will be equal to `dockerImage` config entry. But when the registry is not dockerhub then, the image will be `dockerRegistry + dockerImage`. But the whole pipeline only refer `dockerImage` which works only when pulling from default registry. e.g [here](https://github.com/adoptium/ci-jenkins-pipelines/blob/e20869f23f9941e97a724aa7e28af9c1551ea78d/pipelines/build/common/openjdk_build_pipeline.groovy#L2072)

This change tries to add a shortened tag equal to `dockerImage` to let other calls referring `dockerImage` to work without issue.

sxa (approved):
> When we pull image, if the registry is dockerhub, then local image name will be equal to dockerImage config entry. But when the registry is not dockerhub then, the image will be dockerRegistry + dockerImage. 

Yes that makes sense. Thanks for the explanation - if you pull from another public repo you could just prefix the image with the registry host but since you need authentication that isn't possible.
LGTM.

Issue #1082, opened 2024-07-23
riscv64 builds (or any using docker image tag or digest) broken


andrew-m-leonard:
PR: https://github.com/adoptium/ci-jenkins-pipelines/pull/1076 added a patch to tag using the docker_image to handle scenario when using a dockerRegistry. However, it searches the image ls using the docker_image, which may contain a :tag or @digest, which will not match...
eg: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk23/job/jdk23-linux-riscv64-temurin/9/console
```
10:11:32  + docker image ls
10:11:32  + grep adoptopenjdk/ubuntu2004_build_image:linux-riscv64
10:11:32  + head -n1
10:11:32  + awk {print $1}
[Pipeline] sh
10:11:34  + docker tag  adoptopenjdk/ubuntu2004_build_image:linux-riscv64
10:11:34  Error parsing reference: "" is not a valid repository/tag: invalid reference format
```

docker ls returns the "Repository", you can't locate the entry you need from that list using the plain DOCKER_IMAGE string

sxa:
Took me a bit of time to understand the specific problem here, but for historic reference it is because `docker image ls` gives lines like this:
```
adoptopenjdk/ubuntu2004_build_image           linux-riscv64   d7f86236b5fd   5 days ago       3.01GB
```
which the grep with the colon separation between image and tag will not pick up.

PR #1083, opened 2024-07-23
Move dockerRegistry tag patch into with dockerRegistry section to avoid build break with image tags

andrew-m-leonard:
Fixes: https://github.com/adoptium/ci-jenkins-pipelines/issues/1082

A patch to fix the above build break, fix https://github.com/adoptium/ci-jenkins-pipelines/pull/1076 was added to handle dockerRegistry and adding a local tage of DOCKER_IMAGE, but does not correctly handle the fact the image may have a :tag or @digest

This patch moves that code to the dockerRegistry section, but the proper fix for https://github.com/adoptium/ci-jenkins-pipelines/pull/1076 should be to not do that tagging, but instead probably to use context.docker.withRegistry when necessary to run the container, but that needs proper investigation.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -2056,6 +2056,8 @@ class Build {
                                                     context.docker.image(buildConfig.DOCKER_IMAGE).pull()
                                                 }
                                             }
+                                            def long_docker_image_name = context.sh(script: "docker image ls | grep ${buildConfig.DOCKER_IMAGE} | head -n1 | awk '{print \$1}'", returnStdout:true).trim()
+                                            context.sh(script: "docker tag '${long_docker_image_name}' '${buildConfig.DOCKER_IMAGE}'", returnStdout:false)
                                         } else {
                                             if (buildConfig.DOCKER_ARGS) {
                                                 context.sh(script: "docker pull ${buildConfig.DOCKER_IMAGE} ${buildConfig.DOCKER_ARGS}")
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

andrew-m-leonard:
@mahdipub fyi

sxa (commented):
I'm struggling a little to understand the fix here so just to clarify - is this fix going to work on the assumption that things from a separate registry (i.e. with `DOCKER_REGISTRY` in the pipeline config) won't have a tag or is it more subtle than that?

andrew-m-leonard:
> I'm struggling a little to understand the fix here so just to clarify - is this fix going to work on the assumption that things from a separate registry (i.e. with `DOCKER_REGISTRY` in the pipeline config) won't have a tag or is it more subtle than that?

Correct, the problem @mahdipub was trying to fix was when using a dockerRegistry rather than the default dockerhub, so for his purpose the fix only needs to be when DOCKER_REGISTRY is used

andrew-m-leonard:
The proper fix I suspect is to re-work the pipeline code to correct use the docker context with a specific Registry

andrew-m-leonard:
> I'm struggling a little to understand the fix here so just to clarify - is this fix going to work on the assumption that things from a separate registry (i.e. with `DOCKER_REGISTRY` in the pipeline config) won't have a tag or is it more subtle than that?

If they have a tag, then you are right that fix will not work either..... but I am guessing @mahdipub semeru does not use "tags" ?

sxa (approved):
> Correct, the problem @mahdipub was trying to fix was when using a dockerRegistry rather than the default dockerhub, so for his purpose the fix only needs to be when DOCKER_REGISTRY is used

Gotcha - thanks for confirming

steelhead31 (approved):
LGTM

PR #1084, opened 2024-07-30
Remove use of --user-openjdk-build-root-directory now JDK-8326685 backported

andrew-m-leonard:
JDK-8326685 provides an official fix for the Linux debug map reproducible build issue, this is now backported to jdk-21.0.4, and the workaround to use a user build folder is not required.
The workaround was incorrectly applied to Windows, which then caused a subsequent non-reproducible issue due to relative path building not being used.

Fixes https://github.com/adoptium/temurin-build/issues/3790

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1522,7 +1522,7 @@ class Build {
             def openjdk_build_dir
             def openjdk_build_dir_arg
 
-            # Build as default within OpenJDK src tree, necessary for Windows reproducible builds, due to relative paths
+            // Build as default within OpenJDK src tree, necessary for Windows reproducible builds, due to relative paths
             build_path = 'workspace/build/src/build'
             openjdk_build_dir =  context.WORKSPACE + '/' + build_path
             openjdk_build_dir_arg = ""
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

karianna (approved):

steelhead31 (approved):
Looks ok.

sxa (approved):
Noting that the logic being removed was originally added in https://github.com/adoptium/ci-jenkins-pipelines/pull/998/files
I presume this is not a Windows-specific change given that [JDK-8326685](https://bugs.openjdk.org/browse/JDK-8326685) appears to only mention Linux?

Also, is it still possible to reproduce the latest JDk21/22 releases on Windows if this change isn't included (i.e. with the options included in the July SBOMs)?

andrew-m-leonard:
> Noting that the logic being removed was originally added in https://github.com/adoptium/ci-jenkins-pipelines/pull/998/files I presume this is not a Windows-specific change given that [JDK-8326685](https://bugs.openjdk.org/browse/JDK-8326685) appears to only mention Linux?
> 
> Also, is it still possible to reproduce the latest JDk21/22 releases on Windows if this change isn't included (i.e. with the options included in the July SBOMs)?

It was introduced in this PR https://github.com/adoptium/ci-jenkins-pipelines/commit/e297546378b5fbdb676223eb1ae2a0abe7406679
the one you refer to was a builds break fix.
Yes, it should never have been used for Windows

Yes, reproducing July Windows using the release jobs and July v2024.07.01 branch will work fine.

PR #1085, opened 2024-07-31
Now jdk-22.0.2 is GA and version is STS, we can disable the build pipeline

andrew-m-leonard:
Now jdk-22.0.2 is GA and version is STS, we can disable the build pipeline


andrew-m-leonard:
run tests

sxa (approved):

Issue #1092, opened 2024-08-08
disable build job default for ENABLE_REPRODUCIBLE_COMPARE


andrew-m-leonard:
ENABLE_REPRODUCIBLE_COMPARE is currently defaulting to True for jdk17u and jdk21u, however since we have now moved to a new aqa test "Rebuild_Same_JDK_Reproducibility_Test_0", the build job re-build is now a "duplicate" and can be annoying if you run a test build job and forget to set it to False.

Also, the build pipeline boolean setting when False, does not seem to propagate to the disable the ENABLE_REPRODUCIBLE_COMPARE for the build job correctly.

PR #1093, opened 2024-08-08
Disable the reproduce_compare jobs now aqa-test Rebuild_Same_JDK_Reproducibility_Test_0 running

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/1092

Disable ENABLE_REPRODUCIBLE_COMPARE by default, now we have Rebuild_Same_JDK_Reproducibility_Test_0 in aqa-tests.

Also fix the base pipeline groovy to correctly honour the "Enable Reproduce Compare" option, which mean if enabled for the given platform then enable for that build job... (hence why param description says "might run...")

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -154,6 +154,7 @@ class Builder implements Serializable {
             adjustedScmReference = scmReference - ('_adopt')
         }
 
+context.echo "enableReproducibleCompare = "+enableReproducibleCompare
         return new IndividualBuildConfig(
             JAVA_TO_BUILD: javaToBuild,
             ARCHITECTURE: platformConfig.arch as String,
@@ -897,7 +898,6 @@ class Builder implements Serializable {
                         context.stage(configuration.key) {
                             // Triggering downstream job ${downstreamJobName}
 
-context.echo "config.enableReproducibleCompare = "+config.enableReproducibleCompare
                             def buildJobParams = config.toBuildParams()
 context.echo "buildJobParams = ${buildJobParams}"
 
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -160,7 +160,6 @@ class Builder implements Serializable {
             adjustedScmReference = scmReference - ('_adopt')
         }
 
-context.echo "enableReproducibleCompare = "+configEnableReproducibleCompare
         return new IndividualBuildConfig(
             JAVA_TO_BUILD: javaToBuild,
             ARCHITECTURE: platformConfig.arch as String,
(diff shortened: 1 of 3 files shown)


andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

pipelines/build/common/build_base_file.groovy
```diff
@@ -127,7 +127,13 @@ class Builder implements Serializable {
         if (additionalBuildArgs) {
             buildArgs += ' ' + additionalBuildArgs
         }
-        def enableReproducibleCompare = isEnableReproducibleCompare(platformConfig, variant)
+
+        def configEnableReproducibleCompare = false
+        if (enableReproducibleCompare) {
```
sxa, line 132: I'm not an expert on this code, but this seems to be using `enableReproducibleCompare` when the definition of that is being removed. Have I missed something?

sxa: Thanks - that explanation makes it clear :-)

sxa (approved):

pipelines/build/common/build_base_file.groovy
```diff
@@ -127,7 +127,13 @@ class Builder implements Serializable {
...
+            // Pipeline parameter "enableReproducibleCompare" requests reproducibleCompare if enabled for this platform
+            configEnableReproducibleCompare = isEnableReproducibleCompare(platformConfig, variant)
+        }
+
```
sophia-guo, line 136: I think no need to add a new variable. No need to define a local one either.  The following should work

```
if (enableReproducibleCompare) {
            // Pipeline parameter "enableReproducibleCompare" requests reproducibleCompare if enabled for this platform
            enableReproducibleCompare = isEnableReproducibleCompare(platformConfig, variant)
        }
```

pipelines/build/common/build_base_file.groovy
```diff
@@ -190,7 +196,7 @@ class Builder implements Serializable {
             WEEKLY: isWeekly,
             PUBLISH_NAME: publishName,
             ADOPT_BUILD_NUMBER: adoptBuildNumber,
-            ENABLE_REPRODUCIBLE_COMPARE: enableReproducibleCompare,
+            ENABLE_REPRODUCIBLE_COMPARE: configEnableReproducibleCompare,
```
sophia-guo, line 199: This change should be no needed either.

pipelines/build/common/build_base_file.groovy
andrew-m-leonard: yep looks good

pipelines/build/common/build_base_file.groovy
andrew-m-leonard: yep

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
Only failure is jdk8u Alpine, which we know has an upstream build break.

andrew-m-leonard:
@sophia-guo should be good now

PR #1099, opened 2024-08-19
Windows docker fixes/workarounds

sxa:
A few changes.
Follow up to https://github.com/adoptium/infrastructure/pull/3702
Fixes https://github.com/adoptium/infrastructure/issues/3713
1. Docker SHA for the build image parameter is now set to `{{.id}}` since `{{.RepoDigest}}`' is not available in our case (possibly due to it being a local image, but a quick check suggests that the same problem doesn't look like it exists on Linux/s390x)
2. Workspace is set "safe" via `git config --global safe.directory` otherwise it objects to it being owned by the user on the host and will not complete.


sxa:
run tests

karianna (approved):

PR #1100, opened 2024-08-22
jdk21+ promote windows aarch64 to release pipeline

gdams:


pipelines/jobs/configurations/jdk21u_evaluation.groovy
```diff
@@ -1,7 +1,7 @@
 targetConfigurations = [
-        'aarch64Windows' : [
-                'temurin'
-        ]
+        // 'aarch64Windows' : [
```
karianna, line 2: Not just remove this?

pipelines/jobs/configurations/jdk22u_evaluation.groovy
```diff
@@ -1,7 +1,7 @@
 targetConfigurations = [
-        'aarch64Windows' : [
-                'temurin'
-        ]
+        // 'aarch64Windows' : [
```
karianna, line 2: same

pipelines/jobs/configurations/jdk23_evaluation.groovy
```diff
@@ -1,7 +1,7 @@
 targetConfigurations = [
-        'aarch64Windows' : [
-                'temurin'
-        ]
+        // 'aarch64Windows' : [
```
karianna, line 2: same

pipelines/jobs/configurations/jdk24_evaluation.groovy
```diff
@@ -1,7 +1,7 @@
 targetConfigurations = [
-        'aarch64Windows' : [
-                'temurin'
-        ]
+        // 'aarch64Windows' : [
```
karianna, line 2: same

karianna (changes requested):

pipelines/jobs/configurations/jdk21u_evaluation.groovy
gdams: I figured leaving the block was useful as an example for people who want to add new platforms otherwise it's just empty but I'm also happy to remove it. Note that the comments are quite common in other files. E.g https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/jobs/configurations/jdk11u_evaluation.groovy#L9

karianna: Would probably prefer a permanent comment block at the top explaining how the section is used in that case, looks deliberate otherwise.

karianna (changes requested):

tellison (approved):
Happy with changes following Martijn's suggestion.

sxa (commented):
I'd personally be happier holding off on this until we've verified that all of the test suites are definitely passing in the Temurin builds before merging this.
(It sounded from yesterday's PMC call that we're not quite there yet)

karianna:
> I'd personally be happier holding off on this until we've verified that all of the test suites are definitely passing in the Temurin builds before merging this. (It sounded from yesterday's PMC call that we're not quite there yet)

I know we have them passing when we run aqa-test on MSFT Builds of OpenJDK so the gap should be closeable.

sxa:
> > I'd personally be happier holding off on this until we've verified that all of the test suites are definitely passing in the Temurin builds before merging this. (It sounded from yesterday's PMC call that we're not quite there yet)
> 
> I know we have them passing when we run aqa-test on MSFT Builds of OpenJDK so the gap should be closeable.

Absolutely :-) I don't anticipate any blocking issues but want to ensure that our infrastructure is able to replicate the MSFT results before the promotion occurs. And I'll be very happy when we see this platform in our set of production builds.

karianna (approved):

PR #1103, opened 2024-09-03
WIP: Trial using `context.bat` instead of `context.sh` on Windows in docker

sxa:
This is very much a 'work-in-progress' / feasibility study with the intention of diagnosing the problem described in https://github.com/adoptium/infrastructure/issues/3714. The PR is not ready for review (since it's mostly debug information and experiments) but I'm creating it in the spirit of openness to go alongside the issue :-)

I have attempted to put comments against all of the `context.sh` operations which are performed inside the containers during a docker build to identify which ones are causing problems.

I have converted some of them to `context.bat` based on some of the trial runs (Noting that this cannot be a final solution as-is as it won't work on non-Windows systems unless an `if windows` block is used to switch between them).

The four-digit numbers in the `battable` lines are the line number in the `openjdk_build_pipeline.groovy` and the `windbld` numbers are the job runs of windbld which showed the problem, so are instances where we have definitely hit an issue with those.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1658,7 +1658,8 @@ class Build {
                                     context.withEnv(['BUILD_ARGS=' + signBuildArgs]) {
                                         context.println 'Building an exploded image for signing'
                                         // windbld#254
-                                        context.bat(script: "bash -c 'curl https://ci.adoptium.net/userContent/windows/openjdk-cached-workspace.tar.gz | tar -C /cygdrive/c/workspace/openjdk-build -xpzf -'")
+                                        // context.bat(script: "bash -c 'curl https://ci.adoptium.net/userContent/windows/openjdk-cached-workspace.tar.gz | tar -C /cygdrive/c/workspace/openjdk-build -xpzf -'")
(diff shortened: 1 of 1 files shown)


sxa:
run tests

sxa:
@andrew-m-leonard FYI

This is really horrible, but it's to resolve issues with using `context.sh` in a docker container. I've trapped the situations that are occurring more often - I may need to do more but it's been fairly reliable since I got this far.

LMK what you think of this  (Mostly the `if windows && docker use context.bat else context.sh` bits) - I'd love to hear other suggestions - would it work as a wrapper function for example?

The other changes in this WIP are pretty much just debug or downloading a workspace tarball to avoid me re-running the build each time to see how well it's working (about 7 minutes vs over 2 hours) so you can ignore those bits - obviously they won't be there if this gets to a stage where we're comfortable taking it out of draft.

sxa:
run tests

sxa:
Noting that while this branch is still being used for testing, it will likely get merged into a second PR to cover the refactoring described in https://github.com/adoptium/infrastructure/issues/3709#issuecomment-2333827072

sxa:
Will be integrated into https://github.com/adoptium/ci-jenkins-pipelines/pull/1117

PR #1115, opened 2024-09-23
enable reproducible compare tests on ppc64le and aarch64

### LINKED ISSUE #1114: Enable reproducible compare tests on PPL64le and Aarch64
Enable reproducible compare tests on PPL64le and Aarch64 so we can get instant test results per tag

sophia-guo:
Close #1114


sophia-guo:
Add special.system category for nightly, which is the category tests belong to.

PR #1120, opened 2024-10-03
Evaluation config can be empty for beta trigger if none for a version

andrew-m-leonard:
jdk21u does not have any evaluation targets now win aarch64 has been promoted, but trigger was assuming there should always be one...

Changed evaluation trigger to not throw error if no evaluation targets


karianna (approved):

steelhead31 (approved):
Looks ok.

PR #1122, opened 2024-10-08
Add aarch64Windows arch to filename mapping to daily slack status tool

andrew-m-leonard:
With the Windows aarch64 now promoted to main builds, we need the aarch64Windows -> aarch64_windows asset mapping adding to the daily Slack tool.


steelhead31 (approved):
Looks ok.

gdams (approved):

PR #1135, opened 2024-11-05
Limit docker label to linux special.system

sophia-guo:
Related https://github.com/adoptium/temurin-build/issues/4025


karianna (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -433,7 +435,7 @@ class Build {
                             DYNAMIC_COMPILE = true
                         }
                         def additionalTestLabel = buildConfig.ADDITIONAL_TEST_LABEL
-                        if (testType  == 'dev.openjdk' || testType  == 'special.system') {
+                        if (testType  == 'dev.openjdk' || (testType  == 'special.system' && jobName.contains('linux'))) {
```
smlambert, line 438: Wondering if this adversely affects others using our pipelines (this is specific to Temurin (Eclipse Adoptium 'vendor' specific).  Since no one has complained about this already, I will assume no, but future enhancement would be to frame it to only affect if vendor == Eclipse Adoptium).

smlambert (approved):
Adding a comment for future consideration only, does not block merging this PR.

pipelines/build/common/openjdk_build_pipeline.groovy
sophia-guo: Updated with buildConfig.VARIANT == 'temurin' with comment.

PR #1117, opened 2024-09-26
Refactor signing/assemble code in openjdk_build_pipeline.groovy to support Windows/docker builds

sxa:
There's a lot in here, so apologies in advance to the reviewers who will need a bit of time to review this properly ;-) Noting that there are some `SXAEC:` references in here which are things that still need to be cleaned up but the overall architecture and flow is now functional as per the builds in https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/windbld which are now getting through all the steps and running tests.

Some things done in this PR:
- A new `batOrSh` function that runs commands under the windows batch scripts if we're on windows instead of always using `bash` as this was causing some instability (issue 3714 listed below)
- Split out some parts of buildScripts into separate functions: `buildScriptsAssemble` and `buildScriptsEclipseSigner`
- Each of those two are now listed as a separate "stage" in the pipelines, so we can see how long each part takes. The original `sign` phase has been renamed as `sign tgz/zip` to avoid ambiguity. I'm tempted to change `assemble` to `assemble & SBoM` (Noting that it also includes the `du -k` of the output from `make-adopt-build-farm.sh` which is slow - at least 10 minutes - on my current test machine)
- There are a few `openjdk_build_pipeline:` eyecatchers added so we can more easily find the separate sections while looking at the build logs (`curl -s` the `consoleText` link and grep it for `_pipeline:` - similar to the `build.sh:` lines from the temurin-build repo)
- The setting up of `envVars` has been extracted into the top level `build()` and passed into the `buildScripts()` and `buildScriptsAssemble()` functions that use it
- There is an 'ls -l' of various directories which forces shortname creation. It's a bit of a hack but the shortnames in containers is a bit hit or miss and this makes it reliable ... Without it it won't find various bits of the compiler tools
- I've left `openjdk_build_dir_arg` (obsoleted in https://github.com/adoptium/ci-jenkins-pipelines/pull/1084) commented out but I'm still tempted to just remove it now ...
- Initial stash of files for the internal signing phase now restricts the number of files copied across as `**/*` was unnecessary.

There are still some issues with the workspace cleaning options which may need to be addressed, although that can be done in subsequent PRs. e.g. `rm -rf c:/workspace/openjdk-build/cyclonedx-lib c:/workspace/openjdk-build/security` and `rm -rf ' + context.WORKSPACE + '/workspace/target/*`. Some additional work will be needed before the clean options will work. Ref: https://github.com/adoptium/infrastructure/issues/3723

[...]

PR #1137, opened 2024-11-07
Skip internal signing phase unless on windows/mac

sxa:
The implementation of the internal signing phase which was adjusted in https://github.com/adoptium/ci-jenkins-pipelines/pull/1117 ([issue](https://github.com/adoptium/infrastructure/issues/3286)) caused that phase to be executed on more platforms than is necessary.

There are two fixes in here:
- `trim()` the output from `ls` when setting `build_path` so that it doesn't have a newline character on it (This was already done for windows but not on mac, but is required there too).
- Ensure that the internal signing + assemble phases are only executed on macos and windows. It was trying to run on Linux platforms and tripping up on the stash operations

Notes / alternative options:
- We could have a separate build flag for the internal signing instead of using `ENABLE_SIGNER` for both signing situations instead of doing this on a platform basis (probably a good idea in the future)
- This has been tested on Linux and macos and appears to do what is expected. Windows will take a little longer due to build times, but since macos looks ok this is likely safe to merge now.
- signing issues are not caught during the PR testing process. We could perhaps look at incorporating that in the future.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1520,6 +1520,7 @@ class Build {
         def base_path
         base_path = build_path
         def repoHandler = new RepoHandler(USER_REMOTE_CONFIGS, ADOPT_DEFAULTS_JSON, buildConfig.CI_REF, buildConfig.BUILD_REF)
+
         context.stage('internal sign') {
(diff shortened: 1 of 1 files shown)


steelhead31 (approved):
Looks good.

smlambert (approved):

PR #1140, opened 2024-11-12
Do not archive before assembly phase when internal signing (Fixes cross-compile situation)

sxa:
Fixes this failure condition on windows/aarch64:
```
10:54:05  Skipping writing incomplete metadata for now - will be done in the assemble phase instead
[Pipeline] timeout
10:54:05  Timeout set to expire in 3 hr 0 min
[Pipeline] {
[Pipeline] echo
10:54:05  [INFO] Archiving JSON Files...
[Pipeline] archiveArtifacts
10:54:05  Archiving artifacts
10:54:13  ‘workspace/target/*.json’ doesn’t match anything: ‘workspace’ exists but not ‘workspace/target/*.json’
```
Introduced by the refactor at https://github.com/adoptium/ci-jenkins-pipelines/pull/1117

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1981,32 +1981,35 @@ def buildScriptsAssemble(
                 }
                     
             } finally {
-            
-                // Always archive any artifacts including failed make logs..
-                try {
-                    context.timeout(time: buildTimeouts.BUILD_ARCHIVE_TIMEOUT, unit: 'HOURS') {
-                        // We have already archived cross compiled artifacts, so only archive the metadata files
-                        if (buildConfig.BUILD_ARGS.contains('--cross-compile')) {
-                            context.println '[INFO] Archiving JSON Files...'
-                            context.archiveArtifacts artifacts: 'workspace/target/*.json'
-                        } else {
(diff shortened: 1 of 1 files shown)


sxa:
run tests

sxa:
My latest tests have managed to run through correctly although tripped up on the installer generation phase due to a mismatch between the version built (21.0.5) and the latest version in the repo (21.0.6).

sxa:
Note to reviewers:
- Suggest reviewing with whitespace ignored: https://github.com/adoptium/ci-jenkins-pipelines/pull/1140/files?diff=split&w=1
- Ref the "this is not strictly requird" comment I've left that condition in because the same code is used elsewhere in the file and I'd rather they didn't diverge (an alternate solution would be to move it into a separate function and if that is desired I suggest raising another issue as I really want this tested version in before the EA pipelines kick off tonight)

adamfarley (approved):
LGTM

andrew-m-leonard (approved):
looks good

sxa:
I had [one failure in one of my tests](https://ci.adoptium.net/job/build-scripts/job/utils/job/cross-compiled-version-out/941/console) where the windows/aarch64 machine failed to talk to github during the "cross-compiled-version-out" job but that is assumed to be independent of this PR, so merging so that it can be used during the EA builds tonight.

Issue #1132, opened 2024-10-31
Enhance the build reproducibility healthcheck scripting to identify pipelines per platform


adamfarley:
**Summary**

Currently, we identify the earliest pipeline job that matches a given tag, and assume that is the correct pipeline to use for reproducibility stats across all platforms.

This issue covers the work needed to adopt a per-platform approach, so a user can rerun a build for a specific platform and have the reproducibility stats for that platform be retrieved from the new pipeline job, just as stats for the other platforms are retrieved from another pipeline.

**Details**

In short, this is a continuation of https://github.com/adoptium/ci-jenkins-pipelines/pull/1113

Especially Andrew's comment here: https://github.com/adoptium/ci-jenkins-pipelines/pull/1113/files#r1794987902

**Additional subtask**

When a build is in progress, we get the reproducibility percentages from the in-progress tag (status\['upstreamTag'\]) rather than the previous tag (releaseName). When the former does not contain the latter, we need to get the pipeline build URL and build ID (TRSS) for the latter, and get the reproducibility percentages based on *that*.

adamfarley:
I've been working on this, and my progress can be seen here:

https://github.com/adoptium/ci-jenkins-pipelines/compare/master...adamfarley:ci-jenkins-pipelines:platform_specific_repro_healthcheck

Also, I'm testing my work here: https://ci.adoptium.net/job/adam_temp_job_2/6/

PR #1146, opened 2024-11-19
Platform-specific reproducibility healthcheck

adamfarley:
This change allows us to identify the latest build of a specific tagged jdk source version per platform.

Previously, we would gather results from the oldest build pipeline using a specific tagged source version.

This change allows us to re-run specific platforms, and have the latest build override the first one, while still using the builds from the original pipeline for other platforms.

This change also includes making all of our wget calls more robust, in case we use invalid key parameters for the trss api and become hung.

Resolves https://github.com/adoptium/ci-jenkins-pipelines/issues/1132

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -209,11 +209,10 @@ echo "D6.3"
                     // - Does the build job for one of our listed platforms contain a successful build job?
                     if (onePipelineBuild.status.equals("Done") && (onePipelineBuild.buildResult.equals("UNSTABLE") || onePipelineBuild.buildResult.equals("SUCCESS"))) {
 echo "D6.4"
-                        if (onePipelineBuild.buildParams.toString() ==~ /ENABLE_TESTS..?\:.?true/)
+                        if (onePipelineBuild.buildParams.toString() ==~ /ENABLE_TESTS..?\:.?true/) {
 echo "D6.5"
                             onePipelinePlatformsMap[onePlatformKey] = onePipelineBuild._id
                         }
-                        
 echo "D7"
                     }
                 }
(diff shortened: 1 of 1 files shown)


adamfarley:
This code has been tested, and here is the output:

image

tools/nightly_build_and_test_stats.groovy
```diff
@@ -147,18 +146,120 @@ def getLatestBinariesTag(String version) {
...
+            for (int k = 0 ; k < platformKeys.size() ; k++ ) {
+                String onePlatformKey = platformKeys[k]
+                String onePlatformValue = platformsList[onePlatformKey]
+                if (!onePlatformValue.isEmpty()) {
+                    continue
+                }
+
+                if (onePipelineBuild.buildName.contains("${jdkVersion}-${platformConversionMap[onePlatformKey][0]}")) {
+                    // - Does the build job for one of our listed platforms contain a successful build job?
+                    if (onePipelineBuild.status.equals("Done") && (onePipelineBuild.buildResult.equals("UNSTABLE") || onePipelineBuild.buildResult.equals("SUCCESS"))) {
```
andrew-m-leonard, line 216: although not ideal, checking buildResult is not quite accurate, as the build job maybe FAILURE if one of the sub-test jobs failed (eg.sanity.openjdk), so it will still have published the JDK artifact

In reality we need to check "Is there a JDK archive in the build artifacts", as if it has for an EA Beta build, then it will have been published...

andrew-m-leonard (changes requested):

tools/nightly_build_and_test_stats.groovy
adamfarley: Ok, I've coded the changes for that [here](https://github.com/adoptium/ci-jenkins-pipelines/commit/0cd13e48b82de3cbc67b31496b05bab178b550c4), and I'm testing it [here](https://ci.adoptium.net/job/adam_temp_job_2/62/console).

andrew-m-leonard (approved):

karianna (approved):

adamfarley:
Thanks @andrew-m-leonard and @karianna 🤓

Issue #1149, opened 2024-11-21
Improve build pipeline publish links to warn of build job failures


andrew-m-leonard:
Arising from retrospective comment: https://github.com/adoptium/temurin/issues/54#issuecomment-2426762664

Improve pipeline publish links in the summary to have a abvious WARNING if the given build job failed.... help potentially publishing a bad JDK by misstake.

sophia-guo:
https://github.com/adoptium/ci-jenkins-pipelines/issues/1148#issuecomment-2491722828 Could we just delete the artifacts for failed build steps and hence do not generate the release link at all?

andrew-m-leonard:
> [#1148 (comment)](https://github.com/adoptium/ci-jenkins-pipelines/issues/1148#issuecomment-2491722828) Could we just delete the artifacts for failed build steps and hence do not generate the release link at all?

I think this is the discussion for the other issue, as to whether we publish regardless, which has been our chosen method to now.
This issue is purely to warn to the fact.

sophia-guo:
For release, there is no reason to publish it.

PR #1150, opened 2024-11-21
Improve build pipeline summary publish link text with build failure warning

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/1149

- Provide a "WARNING build result(FAILURE)" type message next to Publish binary link, to warn user the build job was not successful, and provide link to build job to check...
- Also fixed groovy script from bombing out if no .tap files exist

eg.https://ci.adoptium.net/job/build-scripts/job/release-openjdk17-pipeline/80/
image

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -784,7 +784,7 @@ class Builder implements Serializable {
         def releaseWarning = ''
         if ( jobResult != "SUCCESS" && jobResult != "UNSTABLE" ) {
             // Build was not successful, add warning and link to build job
-            releaseWarning = '<a href=' + jobUrl + '><span style="color:red;">WARNING: build result(<b>' + jobResult + '</b>)</span></a> : '
+            releaseWarning = '<a href=' + jobUrl + '><span style="color:red;">WARNING: pipeline status was <b>' + jobResult + '</b></span></a> : '
         }
         
         def tag = "${javaToBuild}-${timestamp}"
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -783,7 +783,7 @@ class Builder implements Serializable {
         def releaseComment = 'BETA publish'
         def prefixLinkComment = ''
         if ( jobResult != "SUCCESS" && jobResult != "UNSTABLE" ) {
-            prefixLinkComment = '<span style="color:red;">WARNING: job result(<b>' + jobResult + '</b>)</span> : '
+            prefixLinkComment = '<span style="color:red;">WARNING: build result(<b>' + jobResult + '</b>)</span> : '
         }
         
         def tag = "${javaToBuild}-${timestamp}"
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -1026,7 +1026,7 @@ class Builder implements Serializable {
                     // Archive tap files as a single tar file if we have any
                     context.sh """
                         cd ${tarDir}/
-                        if [[ $(ls -l *.tap) ]]; then
+                        if [[ `ls -l *.tap` ]]; then
                           tar -czf ${tarTap} *.tap
                         fi
                     """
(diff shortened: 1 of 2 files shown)


sophia-guo:
Does this mean build job status set as unsuccessful due to test failures will also generate this warning link?

andrew-m-leonard:
> Does this mean build job status set as unsuccessful due to test failures will also generate this warning link?

if not SUCCESS|UNSTABLE, yes

andrew-m-leonard:
run tests

pipelines/build/common/build_base_file.groovy
```diff
@@ -776,11 +776,17 @@ class Builder implements Serializable {
...
-    def publishBinary(IndividualBuildConfig config=null) {
+    def publishBinary(IndividualBuildConfig config=null, String jobResult, String jobUrl) {
         def timestamp = new Date().format('yyyy-MM-dd-HH-mm', TimeZone.getTimeZone('UTC'))
         def javaVersion=determineReleaseToolRepoVersion()
         def stageName = 'BETA publish'
         def releaseComment = 'BETA publish'
+        def releaseWarning = ''
+        if ( jobResult != "SUCCESS" && jobResult != "UNSTABLE" ) {
+            // Build was not successful, add warning and link to build job
+            releaseWarning = '<a href=' + jobUrl + '><span style="color:red;">WARNING: build result(<b>' + jobResult + '</b>)</span></a> : '
```
sxa, line 787: > > Does this mean build job status set as unsuccessful due to test failures will also generate this warning link?
> 
> if not SUCCESS|UNSTABLE, yes

In which case I'd recommend changing the messages to "pipeline result (FAILED|ABORTED)" since "build result" is potentially misleading e.g.

```suggestion
            releaseWarning = '<a href=' + jobUrl + '><span style="color:red;">WARNING: pipeline status was <b>' + jobResult + '</b></span></a> : '
```

andrew-m-leonard: yep agreed, done

pipelines/build/common/build_base_file.groovy
```diff
@@ -1014,17 +1020,20 @@ class Builder implements Serializable {
...
+                    // Archive tap files as a single tar file if we have any
+                    def tapExists = context.sh(script: "ls -l ${tarDir}/*.tap", returnStatus:true)
+                    if (tapExists == 0) {
+                        context.sh """
+                            cd ${tarDir}/
+                            tar -czf ${tarTap} *.tap
+                        """
+                        try {
+                            context.timeout(time: pipelineTimeouts.ARCHIVE_ARTIFACTS_TIMEOUT, unit: 'HOURS') {
+                                context.archiveArtifacts artifacts: "${tarDir}/${tarTap}"
```
sxa, line 1032: Will this step fail if it wasn't able to create a tarball?

sxa (approved):
One comment, but pre-approving on the assumption that the answer to my question is "No" :-)

pipelines/build/common/build_base_file.groovy
andrew-m-leonard: this is the tests tap file, so if there is no JDK tarball, there will be no TAP files... hence why I added the check a few lines up to check if there were any... as previously it bombed out the pipeline with an exit code

andrew-m-leonard: so no is the answer, because we check if one exists.

sophia-guo (approved):

sophia-guo:
I would prefer no links at all if the build jar itself is built unsuccessful.

sxa:
> I personally would prefer no link at all if the build jar itself is built unsuccessful. The main point is if the build jar built unsuccessful with the link it may be published by accident. So if no link enabled there will be no accident at all.

Can you clarify what you mean by "build jar"? Are you describing a situation where the warning message would not be displayed? Or do you think the warning messages isn't obvious enough in general as it still lets you override it (which is the intention here)

sophia-guo:
build jar -->  jdk tarballs. 

> Are you describing a situation where the warning message would not be displayed? 
I prefer if there are errors related with building the jdk tarball process there should not be the publish link at all.

sophia-guo:
> What happens here if the build fails and doesn't produce a tarball? 

For release , no link will be given for this case.

PR #1151, opened 2024-11-28
Set all Temurin windows pipelines to use vs2022 label as that is what they are configured to run with

sxa:
New machines, including the docker provisioned ones, will only have Visual Studio 2022 installed so do not have the earlier labels. Despite this, some of the builds are still looking for the labels relating to earlier versions of Visual Studio [despite using VS2022 now](https://github.com/adoptium/temurin-build/pull/3907). This fixes the labels so they are in sync with what the builds are really using.

### DIFF
--- a/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
@@ -71,7 +71,7 @@ class Config11 {
                 os                  : 'windows',
                 arch                : 'x64',
                 additionalNodeLabels: [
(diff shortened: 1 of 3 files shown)


andrew-m-leonard (approved):

sxa:
> or a 32bit tag added to a different vs2022 machine?

Done - good spot. There are no specific 32-bit machines in our set, so the tag is effectively irrelevant now, as all of the machines are capable of building and testing 32-bit builds.

adamfarley (approved):
LGTM

Issue #1152, opened 2024-12-02
betaTrigger_8ea_arm32Linux does not trigger arm32 build if jdk8u main pipeline has already published


andrew-m-leonard:
If jdk8u main pipeline has already published a given release tag, then betaTrigger_8ea_arm32Linux will not trigger when a new aarch32 port build is published for the same upstream tag.

The trigger_beta_build pipeline script only checks for the presence of the tag, not whether a specific asset has been published currently.

Same potential problem with alpine-jdk8u port, but less likely as based on the same upstream tag.

PR #1153, opened 2024-12-02
For jdk8u aarch32|alpine beta triggers check for existance of a specific asset

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/1152

Update trigger_beta_build.groovy to check for the existence of a specific asset rather than rely purely on the release tag existing.
- x64_linux : Main versions
- arm_linux : jdk8u_aarch32
- x64_alpine-linux : jdk8u_x64_alpine

Also prevent evaluation pipelines incorrectly triggering for jdk8u_aarch32 and jdk8u_alpine.

Note: This change will imply that if the given "arch" above is failing to build, then repeat triggers can occur in that scenario, but then there's obviously a build problem anyway...

### DIFF
--- a/pipelines/build/common/trigger_beta_build.groovy
+++ b/pipelines/build/common/trigger_beta_build.groovy
@@ -114,6 +114,53 @@ def loadTargetConfigurations(String javaVersion, String variant, String configSe
     return targetConfigurationsForVariant
 }
 
+// Verify the given published release tag contains the given asset architecture
+def checkJDKAssetExistsForArch(String version, String release, String variant, String arch) {
+    def assetExists = false
+
+    echo "Verifying ${version} assets in release: ${release}"
+
+    def publishVersion = version
+    // aarch32-jdk8u and alpine-jdk8u published as "jdk8u" tags
+    if (version == "aarch32-jdk8u" || version == "alpine-jdk8u") {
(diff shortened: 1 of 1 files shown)


pipelines/build/common/trigger_beta_build.groovy
```diff
@@ -114,6 +114,47 @@ def loadTargetConfigurations(String javaVersion, String variant, String configSe
     return targetConfigurationsForVariant
 }
 
+// Verify the given published release tag contains the given asset architecture
+def checkJDKAssetExistsForArch(String binariesRepo, String version, String releaseTag, String variant, String arch) {
```
adamfarley, line 118:
```suggestion
def checkJDKAssetExistsForArch(String binariesRepo, String version, String releaseTag, String arch) {
```
To make linter happy, as "variant" is unused.

adamfarley (changes requested):
Linter requests 2 changes. Thoughts?

pipelines/build/common/trigger_beta_build.groovy
andrew-m-leonard: yep removed

adamfarley (approved):
LGTM

steelhead31 (approved):
Looks good.

Issue #1096, opened 2024-08-09
Add a build pipeline parameter to avoid the `docker pull` operation


sxa:
Related: 
- https://github.com/adoptium/temurin-build/issues/3749
- https://github.com/adoptium/infrastructure/issues/3286

In https://github.com/adoptium/ci-jenkins-pipelines/pull/611 I made a change to avoid issuing a `docker pull` to get the dockerfile for RHEL. This is due to the OS being licensed and therefore cannot be redistributed. This was necessary for Linux/s390x where we build on RHEL because CentOS was not available.

The way this was done was to search for `rhel` in the image name, which is suboptimal (and in fact I have created images with `notrhel` in the name elsewhere when I wanted to avoid the pull, such as the windows ones) and was never indented to be a permanent solution. We should add a parameter to the pipelines, or have some other mechanism to do the same, to allow the image not to be "pulled". While we want to have all production images uploaded somewhere (location TBC) this would also allow for more rapid local testing of new images on the machines.

PR #1155, opened 2024-12-03
windows: switch to running builds in docker containers

sxa:
We now have systems which are able to run builds in containers on Windows, similar to what we do for Linux.

Like s390x, these images cannot be publicly available so as a stop-gap until https://github.com/adoptium/infrastructure/issues/3217 I'm using container images pre-created on the machine called `win2022_notrhel_image` specifically (the name including `rhel` is to [block it from trying to issue a `docker pull`](https://github.com/adoptium/ci-jenkins-pipelines/pull/611) but going forward [an option would be good](https://github.com/adoptium/ci-jenkins-pipelines/issues/1096)). 

Builds on the earlier work in:
- https://github.com/adoptium/ci-jenkins-pipelines/pull/1117
- https://github.com/adoptium/ci-jenkins-pipelines/pull/1151

Part of https://github.com/adoptium/infrastructure/issues/3286

### DIFF
--- a/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
@@ -70,6 +70,7 @@ class Config11 {
         x64Windows: [
                 os                  : 'windows',
                 arch                : 'x64',
+                dockerImage         : 'win2022_notrhel_image',
                 additionalNodeLabels: [
                         temurin:    'win2022&&vs2022',
                         openj9:     'win2012&&vs2017',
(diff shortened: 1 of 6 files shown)


sxa:
run tests quick 8,21

sxa:
I've added the labels `build` and `win2022` to [the three active machines](https://ci.adoptium.net/label/dockerBuild&&windows/) to allow the PR tests to run. Note that this would allow builds to be scheduled on these machines when they are not fed the docker parameters which is a risk (particularly before this is merged) but is consistent with what we have on x64 e.g. https://ci.adoptium.net/computer/dockerhost%2Dazure%2Dubuntu2204%2Dx64%2D1/ (This is possibly something we should look at separately across all such places where we use containers for building)

sxa:
Merging this despite the tests not yet passing as they reached the end of the build phases and I want this active for the real EA builds this evening (which won't be able to schedule themselves otherwise). I also don't want to do any additional testing as the nodes are Burstable and I want them to build up some credit for the EA pipelines.

PR #1163, opened 2024-12-11
Update daily Slack status to handle multiple head repo versions jdk24 and jdk25

andrew-m-leonard:
Daily Slack status tool assumed only 1 jdk(head) repo versions, but with stabilization branches we now have 2, eg.jdk-24 and jdk-25


adamfarley (approved):
LGTM

PR #1164, opened 2024-12-13
Slack daily build status does not report latest in-progress build correctly

andrew-m-leonard:
If a pipeline is being re-built due to failures in the previous runs, the Slack status does not report it as in-progress.

The getBuildUrl() function should report the "latest" pipeline, not the "oldest".
The latest running pipeline might also not have tests enabled.


adamfarley (approved):
LGTM

steelhead31 (approved):
Looks ok.

PR #1169, opened 2024-12-19
Special case jdk8u Solaris beta trigger

andrew-m-leonard:
Create a beta trigger job for Solaris required for https://github.com/adoptium/infrastructure/issues/3742#

### DIFF
--- a/pipelines/build/common/trigger_beta_build.groovy
+++ b/pipelines/build/common/trigger_beta_build.groovy
@@ -282,7 +282,11 @@ if (triggerMainBuild || triggerEvaluationBuild) {
     if (triggerMainBuild && mainTargetConfigurations != "{}") {
         if (version == 8 && (mainTargetConfigurations.contains("x64Solaris") || mainTargetConfigurations.contains("sparcv9Solaris"))) {
             // Special case to handle building jdk8u Solaris
-            pipelines["main"] = "build-scripts/jobs/jdk8u/jdk8u-solaris-x64-temurin-simple"
+            if (mainTargetConfigurations.contains("x64Solaris")) {
+                pipelines["main"] = "build-scripts/jobs/jdk8u/jdk8u-solaris-x64-temurin-simplepipe"
+            } else {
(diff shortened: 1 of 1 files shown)


sxa (approved):
LGTM and it'll be obvious if this goes in and doesn't work - one suggestions that you can take or leave as you wish :-)

adamfarley (approved):
LGTM

Also, I noticed that (on line 88) we prioritise the jdk##.groovy file over the jdk##u.groovy file (last I checked, we should always prioritise the "u" file):
```
    configFile = "jdk${javaVersion}${configSet}.groovy"
    def rc = sh(script: "curl --fail -LO ${targetConfigPath}/${configFile}", returnStatus: true)
    if (rc != 0) {
        echo "Error loading ${targetConfigPath}/${configFile}, trying ${targetConfigPath}/jdk${javaVersion}u${configSet}.groovy"
        configFile = "jdk${javaVersion}u${configSet}.groovy"
        rc = sh(script: "curl --fail -LO ${targetConfigPath}/${configFile}", returnStatus: true)
        if (rc != 0) {
            echo "Error loading ${targetConfigPath}/${configFile}"
        }
    }
```

Should we replace that with this?

```
    configFile = "jdk${javaVersion}u${configSet}.groovy"
    def rc = sh(script: "curl --fail -LO ${targetConfigPath}/${configFile}", returnStatus: true)
    if (rc != 0) {
        echo "Error loading ${targetConfigPath}/${configFile}, trying ${targetConfigPath}/jdk${javaVersion}${configSet}.groovy"
        configFile = "jdk${javaVersion}${configSet}.groovy"
        rc = sh(script: "curl --fail -LO ${targetConfigPath}/${configFile}", returnStatus: true)
        if (rc != 0) {
            echo "Error loading ${targetConfigPath}/${configFile}"
        } else {
            echo "${targetConfigPath}/${configFile} loaded successfully"
        }
    }
```

It's an optional inclusion. So long as we only have the u or non-u config files (not both), this *should* never be a problem for us.

adamfarley:
> LGTM and it'll be obvious if this goes in and doesn't work - one suggestions that you can take or leave as you wish :-)

Heya, I don't seem to be able to see any suggestions.

PR #1171, opened 2024-12-20
switch `rhel6||centos6` labels for `sw.tool.glibc.2_12`

sxa:
Fixes https://github.com/adoptium/aqa-tests/issues/5754 - this is a more explicit/generic way of determining which machines that tests should be run on.

(Having said that, one of the two machines we have at Adoptium which matches this is offline and the other is going to be decommissioned so this may not serve too much use in the future :-) )

I've added the "new" label to the machines.

@AdamBrousseau would be good to have your approval on this in case your systems will be impacted by this change - I can revert the openj9 label change if required.


karianna (approved):

karianna:
Waiting for @AdamBrousseau to comment

AdamBrousseau (approved):
This is fine from a Semeru builds perspective. We don't have any machines with the glibc label nor do we have any cent/rhel6 machines left so this shouldn't affect us.

sendaoYan (approved):

PR #1174, opened 2025-01-06
Ensure post build clean occurs for non-mac&win platforms

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4095

The refactor of the internal signing pipeline code was not calling postBuildWSclean for non-Windows&Mac platforms

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -2064,10 +2064,7 @@ def buildScriptsAssemble(
                        }
                        throw new Exception("[ERROR] Build archive timeout (${buildTimeouts.BUILD_ARCHIVE_TIMEOUT} HOURS) has been reached. Exiting...")
                    }
-                   // With the exclusion above this is no longer strictly required
-                   if ( !enableSigner ) { // Don't clean if we need the workspace for the later assemble phase
-                       postBuildWSclean(cleanWorkspaceAfter, cleanWorkspaceBuildOutputAfter)
-                   }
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

sxa (commented):
This is a side effect of having `enableSigner` mean two things:
1. Enable GPG signing
2. Enable the internal signing phase on applicable platforms (mac/windows).

Given the comment against this the `if ( !enableSigner )` really wants to be `if ( !enableInternalSigner )`. This change will potentially clear out the workspace on Windows between the initial build and assemble phase which is not necessarily what we want. Would it make4 sense to switch the guard to something like `if ( !enableSigner && (windows || mac) )`

I thought I'd created an issue somewhere to try and create an explicit separate option for internal signing, which I think would be useful, but I can't find it so maybe I haven't created it yet :-)

andrew-m-leonard:
@sxa Isn't that what this line does already? https://github.com/adoptium/ci-jenkins-pipelines/pull/1174/files#diff-b4bb07e2f448f9ad2986d69423ad0b704dcdc4d217d947f465e6ace3fde7a4b3R2049

adamfarley (approved):
LGTM

sxa (approved):
> @sxa Isn't that what this line does already? https://github.com/adoptium/ci-jenkins-pipelines/pull/1174/files#diff-b4bb07e2f448f9ad2986d69423ad0b704dcdc4d217d947f465e6ace3fde7a4b3R2049

Yes ;-)

Issue #1177, opened 2025-01-08
Enable the auto-rerun for Test jobs with FAILURE status


smlambert:
We have previously enabled auto-rerun for Test jobs with UNSTABLE status (yellow/UNSTABLE state).  There is now the ability to also rerun test jobs that failed hard (red/FAILED state), so we will like to turn it on for both AQAvit and TCK testing.

llxia:
The same change was delivered into both the Semeru and Adoptium pipelines. (For some reason, I cannot cherry-pick from the ibmruntimes repo to adoptium repo). 

https://github.com/adoptium/ci-jenkins-pipelines/pull/1178
https://github.com/ibmruntimes/ci-jenkins-pipelines/pull/263

FYI @AdamBrousseau you might encounter merge conflicts during mirroring. Since both PRs are identical, you can choose to either accept or ignore https://github.com/adoptium/ci-jenkins-pipelines/pull/1178 in case of merging conflicts.

PR #1178, opened 2025-01-08
Enable RERUN_FAILURE for AQA test builds

llxia:
resolves: https://github.com/adoptium/ci-jenkins-pipelines/issues/1177

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -516,6 +516,7 @@ class Build {
                         context.string(name: 'VENDOR_TEST_REPOS', value: vendorTestRepos),
                         context.string(name: 'VENDOR_TEST_BRANCHES', value: vendorTestBranches),
                         context.string(name: 'VENDOR_TEST_DIRS', value: vendorTestDirs),
+                        context.booleanParam(name: 'RERUN_FAILURE', value: true),
                         context.string(name: 'RERUN_ITERATIONS', value: "${rerunIterations}")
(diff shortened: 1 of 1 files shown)


llxia:
@smlambert could you please review? Thanks

sxa (approved):

llxia:
This PR only enables RERUN_FAILURE for AQA test builds. Since Adoptium uses another top-level job `AQA_Test_Pipeline` for JCK, RERUN_FAILURE should be set again. It can be set within https://github.com/adoptium/ci-jenkins-pipelines/blob/ff5336261f13f8b2c266e79eba915a391d4973be/pipelines/build/common/openjdk_build_pipeline.groovy#L564 or just set in `AQA_Test_Pipeline`  job config. I would prefer to set `RERUN_FAILURE` to true as default via `AQA_Test_Pipeline`  job config. This is to avoid any confusion in the future if `AQA_Test_Pipeline` is launched alone.

However, If you prefer to set in the pipeline code, please let me know. I will update the PR to cover in `remoteTriggerJckTests()`.

smlambert:
> just set in AQA_Test_Pipeline job config. I would prefer to set RERUN_FAILURE to true as default via AQA_Test_Pipeline job config. This is to avoid any confusion in the future if AQA_Test_Pipeline is launched alone.
However, If you prefer to set in the pipeline code, please let me know. I will update the PR to cover in remoteTriggerJckTests().

Indeed, I think its best to set it in the job config on the server in question rather than passed in as a parameter.  Thanks!

PR #1183, opened 2025-02-07
Correct EA build status in daily Slack build status when recent builds are GA ones

andrew-m-leonard:
The daily slack EA build status incorrectly reports EA Build status just after GA releases, and also incorrectly determines
jdk8u-aarch32 EA build tags. The reason being because the recent build tag back search doesn't search past the GA tags...

- Corrected the EA asset search to skip GA builds
- Corrected the jdk8u-aarch32 GA tag comparison to work out the correct jdk8u-aarch32-port GA tags
- Corrected reported "build in-progress" status when current published build is the latest but it is missing assets...
- When reporting AQA test status and build is in-progress(Streaming) the completed AQA tests were not being reported for the in-progress build

Slack status from yesterday used this PR: https://adoptium.slack.com/archives/C09NW3L2J/p1738841633827389


sxa (approved):
Happy to rubber stamp this on the basis it seems to work and looks sane. Thanks for Sorting the special case that we had for the rogue solaris "release" :-)

steelhead31 (approved):
Looks good.

PR #1202, opened 2025-03-12
Build jtreg version 7.5.1+1

keithc-ca:
Corrects #1186, which should have built 7.5.1+1 instead of 7.5+1; see [8339238: Update to use jtreg 7.5.1](https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/f644a9511ae32adbbeff77b211f82b8d42bdd62d).

### DIFF
--- a/tools/code-tools/jtreg.sh
+++ b/tools/code-tools/jtreg.sh
@@ -2,7 +2,7 @@
 
 ###################################################################
 # Script to build jtreg test suite harness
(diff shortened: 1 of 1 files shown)


smlambert (approved):

sophia-guo:
@keithc-ca does https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L184 need to update or it's done?

sophia-guo (approved):

PR #1208, opened 2025-04-01
Fix Windows sha256 retrieval

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4139

Windows metadata sha256 context.bat() was missing returnStdOut options and needs processing for bat output.

Test build: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-windows-x64-temurin/271/ -
```
...
```

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1436,11 +1436,12 @@ class Build {
                                             """.stripIndent(), returnStdout: true, returnStatus: false).replaceAll('\n', '')
             } else {
                 context.println "Windows detected - running bat to generate SHA256 sums in writeMetadata"
-                String hash_stdout = context.bat(script: "sha256sum ${file}", returnStdout: true, returnStatus: false)
-                // Windows batch returns stdout of <command invoked>\r\n<sha256sum output>\r\n
-                hash = hash_stdout.split('\r').last().replaceAll('\n', '').split(' ').first()
+                //String hash_stdout = context.bat(script: "sha256sum ${file}", returnStdout: true, returnStatus: false)
+                //// Windows batch returns stdout of <command invoked>\r\n<sha256sum output>\r\n
+                //hash = hash_stdout.split('\r').last().replaceAll('\n', '').split(' ').first()
+                hash = context.bat(script: "@sha256sum ${file}", returnStdout: true, returnStatus: false)
             }
-            context.println "archive sha256 = ${hash}"
+            context.println "archive sha256 = /${hash}/"
 
             data.binary_type = type
             data.sha256 = hash
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1436,8 +1436,9 @@ class Build {
                                             """.stripIndent(), returnStdout: true, returnStatus: false).replaceAll('\n', '')
             } else {
                 context.println "Windows detected - running bat to generate SHA256 sums in writeMetadata"
-                hash = context.bat(script: "sha256sum ${file} | cut -f1 -d' '") // .replaceAll('\n', '')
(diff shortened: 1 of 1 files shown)


karianna (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1436,8 +1436,11 @@ class Build {
                                             """.stripIndent(), returnStdout: true, returnStatus: false).replaceAll('\n', '')
             } else {
                 context.println "Windows detected - running bat to generate SHA256 sums in writeMetadata"
-                hash = context.bat(script: "sha256sum ${file} | cut -f1 -d' '") // .replaceAll('\n', '')
+                String hash_stdout = context.bat(script: "sha256sum ${file}", returnStdout: true, returnStatus: false)
+                // Windows batch returns stdout of <command invoked>\r\n<sha256sum output>\r\n
+                hash = hash_stdout.split('\r').last().replaceAll('\n', '').split(' ').first()
             }
+            context.println "archive sha256 = ${hash}"
```
sophia-guo, line 1443: Is this for debug or on purpose to leave as is?

andrew-m-leonard: @sophia-guo I think this is useful, otherwise the hash is not printed out for each archive type, due to the if further down that only writes the first one only...

sophia-guo (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1436,8 +1436,11 @@ class Build {
                                             """.stripIndent(), returnStdout: true, returnStatus: false).replaceAll('\n', '')
             } else {
                 context.println "Windows detected - running bat to generate SHA256 sums in writeMetadata"
-                hash = context.bat(script: "sha256sum ${file} | cut -f1 -d' '") // .replaceAll('\n', '')
+                String hash_stdout = context.bat(script: "sha256sum ${file}", returnStdout: true, returnStatus: false)
+                // Windows batch returns stdout of <command invoked>\r\n<sha256sum output>\r\n
+                hash = hash_stdout.split('\r').last().replaceAll('\n', '').split(' ').first()
```
sxa, line 1441: I'm a bit confused - what do you mean by "Windows batch" in this comment? I don't recall seeing a situation from the windows command line where sha256sum - which I assume is the cygwin one - would behave like that. If it's more subtle than that then we should adjust this comment.

sxa: Ref my [other comment](https://github.com/adoptium/ci-jenkins-pipelines/pull/1208/files#r2027410485) I see this is specific to the pipeline `bat` operation. You could prevent it by using `@sha256sum` which might keep the operation a bit simpler.

```suggestion
                String hash_stdout = context.bat(script: "@sha256sum ${file}", returnStdout: true, returnStatus: false)
```

Or if you don't want to do that can we change the comment please?
```suggestion
                // "bat" in pipelines returns stdout of <command invoked>\r\n<sha256sum output>\r\n
```

sxa (changes requested):

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: i'll give that a try, thanks

andrew-m-leonard: Test build https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-windows-x64-temurin/269/

andrew-m-leonard: tidier yes, need to split, but should work better yes, thanks

sxa (approved):
Approving this simpler version on the assumption that the latest changes are tested and produce the right output :-)

PR #1216, opened 2025-04-10
updateBuildStatus before query the status

sophia-guo:
Ensure remote build status is refreshed.  Query is more accurate.

https://github.com/adoptium/ci-jenkins-pipelines/pull/1204/files update update interval from 3.6 secs to 1 hours. Do the update before query to avoid possible 1 hours delay.


andrew-m-leonard (approved):
lgtm thanks

PR #1222, opened 2025-04-29
Differentiate weekly and release remote trigger displayname

sophia-guo:


sophia-guo:
@andrew-m-leonard this is the easy way I found to differentiate weekly and release remote trigger. Please advise if there are any other easy way to do it.

andrew-m-leonard (approved):
Seems reasonable to me, thanks

smlambert (approved):

PR #1237, opened 2025-07-01
Fix headless and win64 dev.jck trigger

andrew-m-leonard:
Update remoteTriggerJckTests to fix:
- Headless should not trigger dev.jck
- Win64 was incorrectly setting serial_dev


steelhead31 (approved):
Looks good.

PR #1239, opened 2025-07-04
Accommodate suffixes in aarch64 windows openjdk-build dir name

adamfarley:
This platform can include @2 in that dir name, and I wouldn't rule out other ints, so this covers all possibilities.

Testing: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-windows-aarch64-temurin/78/

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1834,7 +1834,12 @@ def buildScriptsAssemble(
       try {
         // This would ideally not be required but it's due to lack of UID mapping in windows containers
         if ( buildConfig.TARGET_OS == 'windows' && buildConfig.DOCKER_IMAGE) {
-            context.bat('chmod -R a+rwX ' + '/cygdrive/c/workspace/openjdk-build*/workspace/build/src/build/*')
+			def cygwin_workspace = context.WORKSPACE
+			if ( !cygwin_workspace.startsWith("/cygdrive") ) {
+				cygwin_workspace = "/cygdrive/c" + cygwin_workspace.substring(2)
+			}
+			
+            context.bat('chmod -R a+rwX ' + cygwin_workspace + '/workspace/build/src/build/*')
         }
         // Restore signed JMODs
         context.unstash 'signed_jmods'
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1834,7 +1834,7 @@ def buildScriptsAssemble(
       try {
         // This would ideally not be required but it's due to lack of UID mapping in windows containers
         if ( buildConfig.TARGET_OS == 'windows' && buildConfig.DOCKER_IMAGE) {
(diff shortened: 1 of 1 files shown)


karianna:
Is it not a finite set?

adamfarley:
> Is it not a finite set?

So far it's either openjdk-build or openjdk-build@2, so yes. 

However, @2 implies the possibility of an @3 under other circumstances, or even an @4, so I think that making the path case-by-case won't serve us in the long run.

We shouldn't be hard-coding the workspace location anyway, so here's another idea:

```
// Where context.WORKSPACE is C:/workspace/openjdk-build@2
def cygwin_workspace = context.WORKSPACE
if ( !cygwin_workspace.startsWith("/cygdrive") ) {
    cygwin_workspace = "/cygdrive/c" + cygwin_workspace.substring(2)
}
context.bat('chmod -R a+rwX ' + cygwin_workspace + '/workspace/build/src/build/*')
```

I've added the change and rerun the test here: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-windows-aarch64-temurin/76/

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1834,7 +1834,12 @@ def buildScriptsAssemble(
       try {
         // This would ideally not be required but it's due to lack of UID mapping in windows containers
         if ( buildConfig.TARGET_OS == 'windows' && buildConfig.DOCKER_IMAGE) {
-            context.bat('chmod -R a+rwX ' + '/cygdrive/c/workspace/openjdk-build*/workspace/build/src/build/*')
+			def cygwin_workspace = context.WORKSPACE
```
karianna, line 1837: whitespace seems out, otherwise LGTM,

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1834,7 +1834,12 @@ def buildScriptsAssemble(
...
+			if ( !cygwin_workspace.startsWith("/cygdrive") ) {
+				cygwin_workspace = "/cygdrive/c" + cygwin_workspace.substring(2)
```
karianna, line 1839: So we force that drive?

adamfarley: That's the current behaviour, yes. Will add logic to convert first character to lowercase (when first char is not a forward-slash) and use that.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1834,7 +1834,12 @@ def buildScriptsAssemble(
...
+			}
+			
+            context.bat('chmod -R a+rwX ' + cygwin_workspace + '/workspace/build/src/build/*')
```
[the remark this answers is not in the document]
adamfarley: I was uncertain about that too, but we need to avoid scope creep before the release.

My priority here is making this code work. If you think we should investigate why the chmod is this specific, we should raise an issue and add it to the backlog.

karianna (approved):

smlambert (approved):

PR #1235, opened 2025-06-26
Add retrigger public aqa_test_pipeline link

sophia-guo:
Close #1241

PR #1243, opened 2025-07-14
V2025.07.01 - cherry pick #1235

sophia-guo:
cherry pick #1235


sophia-guo:
Same as #1244 , two approvals but merging is blocked.

PR #1269, opened 2025-09-05
Allow dryruns against adopt-only ga candidate build source clones

adamfarley:


pipelines/build/openjdk_pipeline.groovy
```diff
@@ -145,7 +145,9 @@ node('worker') {
         }
 
         // If testenv tag is a "-ga" tag, then resolve to the actual openjdk build tag it's tagging
-        if (jdkBranch.contains("-ga")) {
```
andrew-m-leonard, line 148: This is incorrect, and I think the reason is because the original failing pipeline had the wrong scmReference:
https://ci.adoptium.net/job/build-scripts/job/release-openjdk25-pipeline/1/parameters/
scmReference=jdk-25-dryrun-ga

this should be as will be passed by the "trigger" job...:
scmReference=jdk-25+36_adopt

Without a correct scmReference value the temurin-build will to build the right thing...

andrew-m-leonard (changes requested):

pipelines/build/openjdk_pipeline.groovy
andrew-m-leonard: Let me check the "doc" to see what it says, maybe RELEASING.md is wrong...

andrew-m-leonard: The RELEASING.md details using the trigger job...
I think you started the pipeline manually instead, and used the wrong scmReference...

PR #1290, opened 2025-10-15
Allow pipeline release tag check to handle dryrun-ga of a previous -ga level build tag

andrew-m-leonard:
For jdk25u trying to dryrun a previous -ga tag (jdk-25+36 jdk-25-ga), fails as the tag check logic does not expect another "-ga" tag to exist on the actual build tag, so in this case we have:
- jdk-25+36 tagged with : jdk-25-ga and jdk-25-dryrun-ga

We need to ignore any "-ga" tag and only return real build tags when we resolve the git tag check.

Tested here: https://ci.adoptium.net/job/build-scripts/job/release-openjdk25-pipeline/18/console
```
[INFO] Resolved jdk-25-dryrun-ga to upstream build tag jdk-25+36
[Pipeline] echo
[INFO] scmReference=jdk-25+36 matches with JDK25_BRANCH=jdk-25+36 in testenv.properties in aqa-tests release branch.
```


steelhead31 (approved):
Looks ok.

PR #1293, opened 2025-10-16
Update Docker Build With SHA logic

steelhead31:
Fix issue with selecting specific sha based docker images.


steelhead31:
run tests quick 8,21

sxa (approved):
I'm going to approve this because we've spent two hours thrashing around ideas and it seems to make sense. I hope we've covered all the use cases ;-)

FYI @AdamBrousseau in case there's any risk of this to OpenJ9 - we need this to unbreak our use of two different windows images with a different SHA (jdk8u is separate)

PR #1311, opened 2025-11-06
Adding "NotDone" as a valid TRSS status for incomplete pipelines

adamfarley:
As TRSS seems to be using NotDone status for in-progress pipelines instead of Streaming status.

Testing here: https://ci.adoptium.net/job/nightlyBuildAndTestStats_temurin/2163


sxa (approved):
If Shelley's happy with this then I'm also happy to rubber stamp it :-)

PR #1312, opened 2025-11-07
Temurin EA builds are now done for ga tags

andrew-m-leonard:
Update Daily Build Slack status so it does not skip GA tags, now that we do EA builds of all tags : https://github.com/adoptium/ci-jenkins-pipelines/pull/1283


adamfarley (approved):
LGTM

PR #1316, opened 2025-11-17
Remove Win32 burstable node requirement for v2025.10.01

andrew-m-leonard:
Remove Win32 burtsable node requirement for v2025.10.01


adamfarley (approved):
LGTM

PR #1319, opened 2025-11-20
Win32 jck needs to run on ci.role.test.interactive

andrew-m-leonard:
When "burstable" requirement was removed the Win32 "dev.jck" job ran on any node, as ci.role.test.interactive was only being set for x64.


steelhead31 (approved):
lgtm

Issue #1318, opened 2025-11-19
Build job generation cannot gen from user "config repo"


andrew-m-leonard:
If creating a "user" Jenkins setup using "User" pipeline "Config", it is currently not possible to generate the user pipelines from the user config, as "Adoptium" defaults are always picked up by the generation groovy

PR #1320, opened 2025-11-20
Update pipeline and job generation to check existance of "pipeline" scripts and default to Adopt if no User scripts

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/1318

The problem with the existing logic is it made assumption at some points that the repo pointed at with the pipeline config, also hosted the pipeline build scripts. This PR updates the pipeline and job generation to check existence of "pipeline" scripts and default to Adopt if no User scripts.

Tested successfully on a local Jenkins instance.

**Updates:**
- pipelines/build/regeneration/build_pipeline_generator.groovy - This is the "openjdkNN-pipeline" job generator. Added check for existence of the pipeline script in the checked out User pipeline repo, and if not found defaults to the Adopt ci-jenkins-pipelines repo
- pipelines/build/common/config_regeneration.groovy- This is the "build job" generator script. Added check for existence of the build job kick-off script in the checked out User pipeline repo, and if not found defaults to the Adopt ci-jenkins-pipelines repo
- pipelines/build/openjdk_pipeline.groovy - This is the "openjdkNN-pipeline" script, it will dynamically "load" the pipeline job configuration. It is modified to check the "User" pipeline repo first (and not the "scm", ie.the one associated with the running script), and if not found default to the Adopt ci-jenkins-pipelines repo to load the configuration.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -609,32 +609,26 @@ class Build {
 
         def appOptions="customJtx=${excludeRoot}/jenkins/jck_run/jdk${jdkVersion}/${excludePlat}/temurin.jtx"
         if (configureArguments.contains('--enable-headless-only=yes')) {
-           // Headless platforms have no auto-manuals, so do not exclude any tests
+            // Headless platforms have no auto-manuals, so do not exclude any tests
             appOptions=""
         }
 
-        def targets
-        if ("${platform}" == 'ppc64_aix' || "${platform}" == 'sparcv9_solaris' || "${platform}" == 'x86-64_solaris' ||
-             configureArguments.contains('--enable-headless-only=yes')) {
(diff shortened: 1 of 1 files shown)


adamfarley (approved):
LGTM

andrew-m-leonard:
> Is there some documentation updates we can do in the PR alongside these changes to help anyone who needs to use this functionality?
> 
> Also FYI @AdamBrousseau @mahdipub - this shouldn't affect anythnig you do but just for awareness in case this functionality becomes useful to you.

@sxa It is already here https://github.com/adoptium/ci-jenkins-pipelines/blob/master/docs/UsingOurScripts.md
This just fixes the code to do as this doc suggests, ie. your "User" repo can contain whatever "config" or "scripts", and if not found there will default to looking at Adopt's.
So fyi @AdamBrousseau with this PR, you can have a "User" pipelines repo that is purely "just" pipeline config, ie. the pipeline/jobs/configurations/.. folder, plus the pipelines/defaults.json

PR #1322, opened 2025-11-21
Removing Solaris from the jdk8 build pipeline

adamfarley:
JDK8 Solaris should still be built as part of the simplepipe jobs, but the traditional Solaris jobs don't work and will continue to fail if these entries are left in.

Change made as per the recent retrospective: https://github.com/adoptium/temurin/issues/90#issuecomment-3456677438


sxa (approved):

adamfarley:
Thanks @andrew-m-leonard and @sxa 😎

PR #1326, opened 2025-11-25
add all hotspot variants so they are regenerated for trestle, remove third party variants

sxa:
Add `hotspot` variants to the configuration so those jobs are regenerated and kept up to date with a view to using them in [project trestle](https://github.com/adoptium/adoptium/issues/265).

This is a follow on from the work to enable and test on a subset platforms at https://github.com/adoptium/ci-jenkins-pipelines/pull/1317 which has been shown to work as well as the Temurin pipelines.

This PR also removes most of the third party variants which we are no longer building (I've left them on x64 in case we want to test that our underlying code that supports those still works). We may wish to clarify whether Alibaba or Huawei are utilising our pipelines configurations in any way (I suspect not - OpenJ9 certainly are not as confirmed in the previous PR referenced above)

I was tempted to add in arm32/win32 as platforms for jdk25/26 to allow people using trestle to test on a 32-bit platform, and while that is feasible it would also require a suitable boot JDK to be made available which do not currently have on those platforms.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -706,7 +706,7 @@ class Build {
                         EXTRA_OPTIONS: "${extra_options}",
                         SETUP_JCK_RUN: "${setupJCKRun}"
                     ]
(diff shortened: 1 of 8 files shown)


sxa:
run tests

steelhead31 (approved):
Looks good.

PR #1330, opened 2025-11-28
Remove string indicating no scm tag

adamfarley:
Because the sbom validation code doesn't parse it correctly. It instead expects "null", which is being removed in favor of an empty string [here](https://github.com/adoptium/temurin-build/pull/4304/files#diff-479de0bbd23850873cf19a0dc3bb5f23e5c78ecf457bf6f19747d038930f9a7e).


adamfarley:
Thanks @sxa @karianna and @andrew-m-leonard 😎

Issue #1338, opened 2025-12-09
Request: Report remote job status using standard job status format


adamfarley:
When remote jobs complete, their status is not declared in the same format as most child jobs.

Example: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk25u/job/jdk25u-mac-x64-temurin/9

Most jobs:

```
01:07:45.913  Build [Test_Job_Auto_Gen #44428](https://ci.adoptium.net/job/Test_Job_Auto_Gen/44428/) completed: SUCCESS
```

Problem jobs:

```
06:44:52.454  Current sanity.jck Status: FINISHED Build status: FAILURE Remote build URL: https://ci.eclipse.org/temurin-compliance/job/AQA_Test_Pipeline/10389/
[Pipeline] stage
[Pipeline] { (sanity.jck)
[Pipeline] catchError
[Pipeline] {
[Pipeline] error
[Pipeline] }
06:44:52.500  ERROR: sanity.jck not successful, setting stage result to: FAILURE
```

Standardising should make it easier to scan the output for child job issues.

PR #1339, opened 2025-12-09
Produce standard format child job status output to aid triage

adamfarley:
This allows triagers to search through the output for a standard "completed: F" string to identify child jobs which have had problems.

Resolves: https://github.com/adoptium/ci-jenkins-pipelines/issues/1338

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -2359,6 +2359,7 @@ def buildScriptsAssemble(
                     context.stage("${testTarget}") {
                         setStageResult("${testTarget}", remoteJobStatus)
                     }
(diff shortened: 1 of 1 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -2359,6 +2359,7 @@ def buildScriptsAssemble(
                     context.stage("${testTarget}") {
                         setStageResult("${testTarget}", remoteJobStatus)
                     }
+                    context.println "Build ${testTarget} completed: ${remoteJobStatus}"
```
smlambert: I'd like to mimic the same console output as the test pipelines use, so I can re-use the same code in TRSS to find and report status of these remote triggered jobs.

PR #1342, opened 2025-12-15
Adding useful output to remote test triggers

adamfarley:
This change is useful because it lends context to situations when the remote job trigger fails before it can output any text that might help a user understand where a failure has occurred.

Resolves https://github.com/adoptium/ci-jenkins-pipelines/issues/1343


adamfarley:
In hindsight, this message may be in the wrong place.

Running a test to check.

https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-riscv64-temurin/124/

PR #1351, opened 2026-01-09
WIP: Overriding the default labels for xmac smoke tests

adamfarley:
Because the Orka plugin doesn't accept labels with "or" logic in them, which is the default.


adamfarley:
In normal aqa testing, the "or" logic seems to launch the orka nodes without issue.

Example: https://ci.adoptium.net/job/Test_openjdk11_hs_extended.functional_x86-64_mac/552/

So this change may not be needed. Consulting the team.

adamfarley:
Closing this as resolved via orka template label reconfiguration.

PR #1493, opened 2026-08-18
Avoid corrupt upstreamTag when CSPU release shares GA commit

### LINKED ISSUE #1492: resolveTag is broken as CSPU tags on the exact same commit as the GA release
resolveTag is broken as CSPU tags on the exact same commit as the GA release. 

For example in OpenJDK's jdk17u repository, the CSPUrelease jdk-17.0.20.1+0 was tagged on the exact same commit as the GA release jdk-17.0.20+8 and jdk-17.0.20-ga. This corrupts upstreamTag as`jdk-17.0.20+8jdk-17.0.20.1+0` and could cause pipeline fail.

8cbbca61...   refs/tags/jdk-17.0.20+8^{}
8cbbca61...   refs/tags/jdk-17.0.20-ga^{}
8cbbca61...   refs/tags/jdk-17.0.20.1+0^{}    ← same commit

sophia-guo:
Fix #1492 

Adding `grep "${versionPrefix}+"` to restrict the result to only tags of the exact same version, excluding sibling CPU
release tags that happen to share the same commit.

Assist with BOB

### DIFF
--- a/pipelines/build/openjdk_pipeline.groovy
+++ b/pipelines/build/openjdk_pipeline.groovy
@@ -96,7 +96,10 @@ def resolveGaTag(String jdkVersion, String jdkBranch) {
         def foundRepo = resolveGaCommit.get(0)
         def foundSHA  = resolveGaCommit.get(1)
         // Find upstream target actual tag, ignore any other "-ga" tags, must be a real build tag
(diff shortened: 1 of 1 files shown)


steelhead31 (dismissed):
Looks ok, do validate co-pilots assumption re: JDK8, the proposed fix by copilot would allow the fix to work for both.

steelhead31 (approved):
Looks good.

karianna (approved):

